Issue #21, opened 2020-03-06
[Dev branch] Issues with autometa/common/coverage.py


jason-c-kwan:
This script contains functions to extract coverages from alignments and assemblies, as appropriate.

Issues:

1. The script does not have a very descriptive docstring.

2. In the `if __name__ == '__main__':` block, the argparse setup does not specify that the option `--nproc` (number of processors to use) should be an `int`. 

3. The functions in the script are both able to work out coverages from alignments (`get`) and extract coverages from SPAdes headers in the assembly (`from_spades_names`). However, the `main` function only runs `get`, and there is no way to do SPAdes extraction through the command line.

4. One major problem with the `get` function is that it assumes that we are dealing with paired-end Illumina reads when doing an alignment, and so it would be impossible to use single-end, IonTorrent or long read technology (see Issue #20 ). 

5. In the `get` function, there is a long `if` cascade that basically works out which files exist, and therefore which need to be created. While this is technically correct, it is not very readable (see below). Refactoring or some comments might help. Another problem is that when the function checks whether a stage has been done, all it is checking is whether the file exists, not whether it is empty or whether it is corrupted. There needs to be a better way to verify that these steps have finished correctly, otherwise it will be a nightmare for users to troubleshoot. Alignments etc. are resource intensive and take a long time, so it is reasonable to expect for them to crash or otherwise finish half way through.

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/coverage.py#L128-L148

6. The `make_length_table` function does not have a docstring. Also, the first line is not very readable (could be improved with a docstring):

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/coverage.py#L69

Issue #18, opened 2020-03-05
[Dev branch] Issues with autometa/common/external/bedtools.py


jason-c-kwan:
This is a script which contains wrapper functions for running [https://bedtools.readthedocs.io/en/latest/](Bedtools) and parsing the results.

Issues:

1. The `genomecov` function doesn't return anything, so this should be stated in the docstring.

2. If the `out` path already exists, then the `genomecov` function raises a FileExistsError. However, it would be good to perhaps have a `force` flag so that it will go ahead and overwrite? Otherwise if something goes wrong in the middle of this step, it is difficult to re-run without manually deleting the file. An alternative would be to add code that deletes the output file in the event of `retcode` coming back as not equal to zero.

3. As a first step in the `parse` function, it checks whether the `out` path exists, and if so it will parse as a Pandas dataframe and return that. However, there is no code to check whether the table is valid. Perhaps the error-handling within Pandas is enough?

4. The `parse` function loads the `.bed` file as a Pandas dataframe with the following column names: `contig`, `depth`, `bases`, `length`, `breadth`. However, according to the [https://bedtools.readthedocs.io/en/latest/content/tools/genomecov.html](the Bedtools docs), the last column is the fraction of bases on the chromosome (contig in this case) with depth equal to column 2, so the name `breadth` is perhaps not very descriptive.

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/external/bedtools.py#L102-L103

5. Likewise I am not sure the name `total_breadth` is appropriate for this variable:

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/external/bedtools.py#L107-L109

WiscEvan:
I'm taking this on, as I've found a line that has implications in #21 (which I'm already working on).

WiscEvan:
@jason-c-kwan What would you prefer the column names (`breadth` and `total_breadth` to be?

I was referencing [this](https://www.danielecook.com/calculate-depth-and-breadth-of-coverage-from-a-bam-file/) when naming the columns.

Issue #17, opened 2020-03-05
[Dev branch] Issues with autometa/common/kmers.py


jason-c-kwan:
This file contains functions to count and retrieve k-mers of the given sequences. The functions in this file also handle embedding (right now, TSNE or UMAP). The docstring includes a note of a TODO task - to make a separate file to handle parallel, work-queue processing. I think this refers to the fact that k-mer counting per contig, and also normalization could be parallelized. 

Issues:

1. The `load` function is used to parse a pre-made kmer frequency table, and it checks that the `kmers_fpath` that it is given exists and is nonzero size. Otherwise, an error is raised. However, the problem here is that there are no checks on whether the file exists and is incomplete or corrupted. Should that happen the user will get a cryptic error. Perhaps a function could be added to check the format of the file?

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/kmers.py#L102-L104

2. `revcomp` function. This is a simple function which gives the reverse complement of an input string. However, it basically just assumes that the input will be uppercase, and it cannot handle any of the IUPAC degenerate nucleotide codes. If any weird letters are encountered, a value of `-1` is returned rather than the reverse complement string. Conceptually that is OK, because the function is just used internally to reverse complement k-mers. However, first of all, this should be explicitly stated in a docstring for the function so no one misuses the function, and secondly, elsewhere in the script there is no handling for a `-1` value. So actually, you will end up with counts for a k-mer called `-1` in the following snippet:

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/kmers.py#L76-L80

NOTE: The script also uses the related `reverse_complement` function from BioPython/SeqIO, for example here:

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/kmers.py#L129-L139

This is fine, because even though BioPython presumably can handle degenerate codes, if the returned value is not in the k-mer matrix, then it is skipped (which is expected behavior), especially since the object it is working on is a SeqRecord. However, to avoid a weird error in the future, it might be worth writing comments near instances of `reverse_complement()` that explicitly state that we intend on using the BioPython version there.

[...]

Issue #16, opened 2020-03-05
[Dev branch] Issues with autometa/common/external/prodigal.py


jason-c-kwan:
This file contains functions used to call ORFs and get ORF translations with external program Prodigal.

Issues found:

1. In the `if __name__ == '__main__':` block, none of the argparse arguments have specified types. The problem here is that if you don't specify types, then you loose some in-build error handling from argparse. For example, `--cpus` should be `int` and `--parallel` should be specified as `bool`.

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/external/prodigal.py#L197

2. In the `run` function, it will uncompress the assembly file if the path ends in `.gz`. However, this is space inefficient as it could be achieved by opening the file with the `gzip` module.

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/external/prodigal.py#L69-L76

3. The commands run when the `--parallel` flag is `True` or `False` are subtly different - the non-parallel command uses the `-m` flag for Prodigal, whereas the parallel version does not. According to the Prodigal help message this option treats runs of N as masked sequence; don't build genes across them. So the option should be the same in parallel and non-parallel.

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/external/prodigal.py#L111-L120

4. In both parallel and non-parallel cases, the Prodigal command is run through the `subprocess` module, and if the return code is not zero (i.e. there was an error), then a warning is logged but nothing further is done. The function will check whether the expected output paths `nucls_out` and `prots_out` exist, but this does not protect against the case where Prodigal fails after writing a partial output.

https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/external/prodigal.py#L148-L153

5. The name of the `get_orf_translations` function is misleading. A user might assume that it returns the protein translations of ORFs, but what it is really doing is constructing a dictionary connecting the full gene name (key) to the contig name (value). This is possible because Prodigal tacks on some extra stuff to the contig name for each gene it finds, and explicitly states the gene name in the header line of output fasta files. I suggest that this function is called something else. **NOTE: This function is called in `autometa/common/external/hmmer.py` and `autometa/taxonomy/lca.py`, so make sure not to break anything.

jason-c-kwan:
[...]

PR #25, opened 2020-03-11
Resolving issues #16, #17, #18, #21  and update to Autometa API and Logger

WiscEvan:
# Issue 17: `kmers.py`
## 1. Exception handling for load
https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/kmers.py#L83
Custom exception classes have been written in autometa/common/exceptions.py. Relevant exceptions for `kmers.py` are `IncorrectKmerFormat` and `KmerEmbeddingError`.

#### Exceptions References:
- https://www.programiz.com/python-programming/user-defined-exception
- https://towardsdatascience.com/how-to-define-custom-exception-classes-in-python-bfa346629bca

## 2. & 3. revcomp function in `init_kmers` and `init_kmers` documentation
https://github.com/KwanLab/Autometa/issues/17#L76 The case where -1 will execute should never occur. As `init_kmers` will alway generate the kmers being reverse complemented (`all_kmers`) See https://github.com/KwanLab/Autometa/blob/01a432d97044be6522b969a6495be61aeffa2ba7/autometa/common/kmers.py#L65-L66 using the standard ATCG and never encounter IUPAC degenerate nucleotide codes.

## 4. k-mer counting with multiprocessing
Question - has this been tested to see if it actually uses multiple processors efficiently? Or that it is faster than single processing?
Answer - This has been tested and it appears to run 2-4 times as fast as the sequential counter.

## 5,6,7. Case: k-mer length is longer than record length (handling for all respective counting functions)

Contigs where this case applies are returned with the k-mer counts filled as NA. After counting has finished counts of 0 are converted to Na. The contigs without _any_ counts are now removed in the normalization step prior to the CLR transform. 

## 7(continued) & 8. CLR calculation

This is performed in `df.dropna(axis='index', how='all')`.

- `df.dropna(axis='index', how='all')`; Drop contig if no counts were recovered (handling of edge case described in comment 7 above).
- `df.dropna(axis='columns', how='all')` ; Drop k-mer column if no contigs contain this k-mer
- `how='all'` ; only drop if all values being investigated are Na

Both in conjunction specify to only drop the row if all of the columns are Na. After dropping the appropriate columns, Na values are filled with 0 and the normalization procedure then takes place.

## 9 & 10. `embed` function

- n_components has been changed to default of 2. 
- generic text has been removed from docstring.

## 11. TSNE

I think this may require running the python2.7 TSNE via a separate script specific for this version of TSNE. 
 - Located here -> [python 2.7 bhsne](https://github.com/KwanLab/Autometa/blob/dev/autometa/binning/bhsne.py)
- This is already referenced in issue #2 

# Issue 16: prodigal.py

## 1. Argparse

argparse does not support passing bool as a type arg. However, this is handled in how the booleans were defined. (Pass flags to operate as the switch rather than providing boolean values that need to be interpreted from a string from command line to a boolean)

## 2 gzip 

- Prodigal does not support gzipped files

## 3. prodigal command congruent for parallel and sequential orf calling

- done

## 4. returncode checks

a warning is logged and the files are checked because GNU parallel may return other codes that are warnings but not errors. I.e. difficult to discern based on the return code whether the run was successful.

## 5. `get_orf_translations` function name

Two functions are written for conversion of ORF headers and contig headers:

1. func: `contigs_from_headers` - Retrieve a dict of {ORF_ID: contig_ID} to map ORFs to their respective contigs.
2. func: `orf_records_from_contigs` - Retrieve list of SeqIO.SeqRecord objects of ORF records corresponding to input contigs. The docstring clarifies that this function requires an _already_ prodigal annotated file. Otherwise, I'm not sure what to name this to make it more clear.


autometa/binning/recursive_dbscan.py
```diff
@@ -30,6 +30,7 @@
 
 from autometa.common.markers import Markers
 from autometa.common import kmers
+from autometa.common.exceptions import RecursiveDBSCANError
 # TODO: This should be
 # from autometa.common.kmers import Kmers
 # So later we can simply/and more clearly do Kmers.load(kmers_fpath).embed(method)
```
jason-c-kwan: This comment is now in the wrong place.

autometa/common/coverage.py
```diff
@@ -60,16 +60,33 @@ def from_spades_names(records):
         Why the exception is raised.
```
jason-c-kwan: Item 1 in issue #21 not addressed.

autometa/common/coverage.py
```diff
@@ -60,16 +60,33 @@ def from_spades_names(records):
...
+    str
+        </path/to/lengths.tsv>
+    """
+    seqs = {record.id:len(record) for record in SeqIO.parse(fasta, 'fasta')}
+    lengths = pd.Series(seqs, name='length')
+    lengths.index.name = 'contig'
+    lengths.to_csv(out, sep='\t', index=True, header=True)
     return out
 
 def get(fasta, out, fwd_reads=None, rev_reads=None, sam=None, bam=None, lengths=None,
```
jason-c-kwan: If I'm not mistaken this function still does not handle single end data (issue #21, item 4).

autometa/common/exceptions.py
```diff
@@ -0,0 +1,72 @@
...
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with Autometa. If not, see <http://www.gnu.org/licenses/>.
+
+File containing customized AutometaExceptions for more specific exception handling
+"""
+
+class AutometaException(Exception):
+    """docstring for AutometaException."""
```
jason-c-kwan: The docstring could be more descriptive. For example, describing circumstances in which the error is raised.

autometa/common/exceptions.py
```diff
@@ -0,0 +1,72 @@
...
+
+    issue_request = '''
+    An error was encountered!
+
+    Please help us fix your problem!
+
+    You may file an issue with us at https://github.com/KwanLab/Autometa/issues/new
+    '''
+    def __str__(self):
+        return f'{self.value}\n\n{issue_request}'
```
jason-c-kwan: It seems that in Autometa.py the whole of `main` is in a try/except block, which also prints this message. So if this error is raised, will the user see the message twice?

autometa/common/exceptions.py
```diff
@@ -0,0 +1,72 @@
...
+class KmerFormatError(AutometaException):
+    """KmerFormatError exception class."""
+
+    def __init__(self, fpath):
+        super(AutometaException, self).__init__(fpath)
+        self.fpath = fpath
+
+    def __str__(self):
+        return f'{self.fpath} does not contain a \"contig\" column. '\
+        'Ensure the k-mer matrix was properly generated.'
```
jason-c-kwan: I'm not sure if this is that descriptive to a user. Would it be better to say "There is something wrong with the k-mer matrix file. Perhaps it is corrupted? Try deleting [full path] and trying again".

autometa/common/exceptions.py
```diff
@@ -0,0 +1,72 @@
...
+
+class KmerEmbeddingError(AutometaException):
+    """KmerEmbeddingError exception class."""
```
jason-c-kwan: It would be better to have a more descriptive docstring here and an error message for the user. Under what circumstances is this thrown?

autometa/common/exceptions.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return self.value
+
+class RecursiveDBSCANError(AutometaException):
+    """RecursiveDBSCANError exception class."""
```
jason-c-kwan: Same comment here. From the code I can't tell why this exists and there is no message for the user.

autometa/common/external/bedtools.py
```diff
@@ -90,13 +96,18 @@ def parse(bed, out=None):
...
-        return pd.read_csv(out, sep='\t', usecols=cols, index_col='contig')
+    if out and os.path.exists(out) and not os.stat(out).st_size == 0:
+        try:
+            cols = ['contig','coverage']
+            return pd.read_csv(out, sep='\t', usecols=cols, index_col='contig')
+        except ValueError as err:
+            raise ValueError(f'InvalidTableFormat: {out}')
     if not os.path.exists(bed):
         raise FileNotFoundError(bed)
     names = ['contig','depth','bases','length','breadth']
```
jason-c-kwan: See Issue #18, point 4. Perhaps `depth_fraction` is more descriptive?

autometa/common/external/bedtools.py
```diff
@@ -107,15 +118,15 @@ def parse(bed, out=None):
     df = df.assign(total_breadth=lambda x: x.depth * x.bases)
```
jason-c-kwan: See Issue #18, point 4. Perhaps `depth_product` would be more descriptive? The problem with `total_breadth` is that I can't tell what it is from the name, and it also is not a total of anything for the contig.

jason-c-kwan:
You said that Prodigal doesn't support `.gz` files, but could the file be piped into Prodigal with the unix command `zcat`? I'm thinking of something like: `zcat asm.fasta.gz > prodigal ...`.

autometa/common/external/prodigal.py
```diff
@@ -121,41 +125,50 @@ def run(assembly, nucls_out, prots_out, force=False,cpus=0,parallel=True):
...
+            shutil.rmtree(tmpdir)
     else:
         with open(os.devnull, 'w') as stdout, open(os.devnull, 'w') as stderr:
             proc = subprocess.run(cmd, stdout=stdout, stderr=stderr)
             returncode = proc.returncode
     if returncode:
         logger.warning(f'Args:{cmd} ReturnCode:{returncode}')
+        # COMBAK: Check all possible return codes for GNU parallel
     for fp in [nucls_out, prots_out]:
         if not os.path.exists(fp):
```
jason-c-kwan: See issue #16, point 4. Perhaps BioPython could be used to check that `nucls_out` and `prots_out` are valid fasta files?

autometa/common/kmers.py
```diff
@@ -41,12 +41,27 @@
...
+    Parameters
+    ----------
+    string : str
+        Description of parameter `string`.
+
+    Returns
+    -------
+    str or int(-1)
+        reverse complemented string.
+        Note: If any weird letters are encountered, int value of -1 is returned.
```
jason-c-kwan: Nitpick, but in the comment, I would include that the reason for this is so that we don't count k-mers with weird characters. Otherwise someone in the future might decide that this needs "fixing".

autometa/common/kmers.py
```diff
@@ -101,18 +127,58 @@ def load(kmers_fpath):
     """
```
jason-c-kwan: Docstring should be amended to mention the new KmerFormatError.

autometa/common/kmers.py
```diff
@@ -70,14 +96,14 @@ def init_kmers(kmer_size=5):
             for char in dna_letters:
                 new_list.append(current_seq + char)
         all_kmers = new_list
-    # Now we trim k-mers and put them in the dictionary
```
jason-c-kwan: Might be useful to keep this comment here, minus the question, of course, and perhaps more clear.

autometa/common/exceptions.py
WiscEvan: A message is passed in to the exception. The base AutometaException wrapping functionality is not yet implemented, but the base AutometaException can act as the wrapper for emitting a message to submit a user issue based on the error encountered. If an an exception is encountered but an AutometaException is not emitted, a new issue request can be suggested to the user so we can account for any bugs of which we are not aware.

autometa/binning/recursive_dbscan.py
WiscEvan: I'm not sure I understand the comment. Do you mean to mark this as resolved to avoid confusion?

autometa/common/coverage.py
WiscEvan: The function is 7 lines and (I think) reads almost like a sentence. 

I.e. "for record id in records split cov and take the last element."

 An example header is also located in the description for the split to be interpreted with the example above. The name of the series is specified and the index name is assigned until returned. I'll remove the exception docstring. Otherwise, how else would I make this simpler without being periphrastic?

autometa/common/coverage.py
```diff
@@ -171,15 +224,22 @@ def main(args):
         format='%(asctime)s : %(name)s : %(levelname)s : %(message)s',
         datefmt='%m/%d/%Y %I:%M:%S %p',
         level=logger.DEBUG)
-    parser = argparse.ArgumentParser('Autometa Coverage')
+    parser = argparse.ArgumentParser(description='Constuct contig coverage table given an input assembly and reads.')
```
WiscEvan: I interpreted issue-#21 item 1 as the main script description here.

autometa/common/exceptions.py
WiscEvan: This has not yet been implemented, but the idea here is to have an AutometaException wrapper to demarkate between exceptions of which we are aware and are handling/have handled and others that we have not yet encountered. In the latter case, we can emit a message to the user to request they submit an issue.

WiscEvan: The `AutometaException` would be the base exception class for Autometa and other exceptions would be subclassed from this.

autometa/common/exceptions.py
WiscEvan: Yes, possibly. The user could have also provided the wrong file, in which case we would not like them to delete anything and simply pass in the appropriate file. Hard to say. I think this would be good to emit to the user though. Probably messages accounting for both situations would be best.

autometa/common/exceptions.py
WiscEvan: The `RecursiveDBSCANError` is passed a message and is described here:

https://github.com/WiscEvan/Autometa/blob/5c735796764df761558d5035bfd7248409a5fe1a/autometa/binning/recursive_dbscan.py#L340-L343

autometa/common/kmers.py
WiscEvan: This exception will _never_ be encountered. `revcomp()` is only used once and this is used on the `all_kmers` variable which is initialized in the script. Therefore, perhaps for more clarity, this could be removed entirely.

autometa/common/external/bedtools.py
WiscEvan: I'll switch the column name to `depth_fraction`.

autometa/common/external/bedtools.py
WiscEvan: I'll switch the column name to `depth_product`

autometa/common/coverage.py
WiscEvan: A `kwargs` keyword argument is allowed within the bowtie script.

See:

1. [bowtie2 align function definition](https://github.com/WiscEvan/Autometa/blob/5c735796764df761558d5035bfd7248409a5fe1a/autometa/common/external/bowtie.py#L83-L101 "bowtie align function definition with docstring")
2. [adding additional options](https://github.com/WiscEvan/Autometa/blob/5c735796764df761558d5035bfd7248409a5fe1a/autometa/common/external/bowtie.py#L115)

single-end (as well as other) options can be passed in via `kwargs`

autometa/common/exceptions.py
WiscEvan: > You said that Prodigal doesn't support `.gz` files, but could the file be piped into Prodigal with the unix command `zcat`? I'm thinking of something like: `zcat asm.fasta.gz > prodigal ...`.

This is a feature request for the next version of prodigal, however, it does not appear this is currently available. See discussion [here](https://github.com/hyattpd/Prodigal/issues/63)

Similarly you can try:

`gzip tests/data/metagenome.fna && gzip -c tests/data/metagenome.fna.gz | prodigal -m -o /dev/null -p meta -a test.faa -d test.fna`

This results in genes found in sequences with runs of N's and the resulting files are binary encoded.

A few example output lines from test.fna below:

```shell
>^W<82>^Vˈ<91>.<FD><92>0^\<97><9F>~<EA>Ô<C4><F9>^N^YZ<F4>^^}<E0>_1 # 1 # 5538 # 1 # ID=2_1;partial=11;start_type=Edge;rbs_motif=None;rbs_spacer=None;gc_cont=0.930
NNNGTNNNNNNNNCNNNGNNATNGNGNTNNNNNATNANNNNGNNNNNNNNNNNNNNANNNNNGNNNNNNN
NNNNNNCNNNNNNNNNNNNNNNNNNTNNNANNNNNNNNNNANNCNNGNGNNNNNNNNNNNNNNNNNCNNN
NNNNNNNNNNANNNNNNTNNNNGNNNNNNNNNNNNNNNCNNNNNNNNNCNNNNNNNNNTGNNNNNNNNNN
TNNNNNNNNNNGNNNNNNANNNNNNNNNNNTNNNNGTNNNNNNNNNNNNNNNNNNNNNCNNNNNNNNNNN
NNNANNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNCNNNNGNNNNNNNNNNNNNNNNNNNNNNNNNN
NTNGNNNNNNNCNNNTNCNNNNNNTNNNNNNNGNANANNNANNTTCNNNNNGNNNNGNNNNNNNNANNNN
NNNNNNNNNNNNNNNNNNNNNNNNNNGNNNNANNNACCNNNNNNNNNNNNCNNNNNNNNNCNNNNNNNNN
```

WiscEvan:
> You said that Prodigal doesn't support `.gz` files, but could the file be piped into Prodigal with the unix command `zcat`? I'm thinking of something like: `zcat asm.fasta.gz > prodigal ...`.

This is a feature request for the next version of prodigal, however, it does not appear this is currently available. See discussion [here](https://github.com/hyattpd/Prodigal/issues/63)

Similarly you can try:

`gzip tests/data/metagenome.fna && gzip -c tests/data/metagenome.fna.gz | prodigal -m -o /dev/null -p meta -a test.faa -d test.fna`

This results in genes found in sequences with runs of N's and the resulting files are binary encoded.

A few example output lines from test.fna below:

```shell
>^W<82>^Vˈ<91>.<FD><92>0^\<97><9F>~<EA>Ô<C4><F9>^N^YZ<F4>^^}<E0>_1 # 1 # 5538 # 1 # ID=2_1;partial=11;start_type=Edge;rbs_motif=None;rbs_spacer=None;gc_cont=0.930
NNNGTNNNNNNNNCNNNGNNATNGNGNTNNNNNATNANNNNGNNNNNNNNNNNNNNANNNNNGNNNNNNN
NNNNNNCNNNNNNNNNNNNNNNNNNTNNNANNNNNNNNNNANNCNNGNGNNNNNNNNNNNNNNNNNCNNN
NNNNNNNNNNANNNNNNTNNNNGNNNNNNNNNNNNNNNCNNNNNNNNNCNNNNNNNNNTGNNNNNNNNNN
TNNNNNNNNNNGNNNNNNANNNNNNNNNNNTNNNNGTNNNNNNNNNNNNNNNNNNNNNCNNNNNNNNNNN
NNNANNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNCNNNNGNNNNNNNNNNNNNNNNNNNNNNNNNN
NTNGNNNNNNNCNNNTNCNNNNNNTNNNNNNNGNANANNNANNTTCNNNNNGNNNNGNNNNNNNNANNNN
NNNNNNNNNNNNNNNNNNNNNNNNNNGNNNNANNNACCNNNNNNNNNNNNCNNNNNNNNNCNNNNNNNNN
```

autometa/binning/recursive_dbscan.py
jason-c-kwan: Where it says `TODO: This should be from autometa.common.kmers import Kmers`. The comment refers to line 32, and so is in the wrong place because you inserted line 33 in there.

autometa/common/coverage.py
jason-c-kwan: The description `Autometa coverage` at the beginning of the script does not seem very descriptive to me.

autometa/common/coverage.py
jason-c-kwan: But it throws an error if it has `fwd_reads` but not `rev_reads`...

jason-c-kwan: From the `bowtie.align` definition it looks like it always includes `-1` and `-2`, so it will now work if either fwd or rev are not defined.

autometa/common/kmers.py
jason-c-kwan: Maybe. I just thought it would be worth including a comment about our reasoning. Otherwise it is a bit difficult to grasp from code review alone.

autometa/common/kmers.py
```diff
@@ -245,32 +338,39 @@ def normalize(df):
...
         Reduce k-mer frequencies dimensions to `pca_dimensions` (the default is 50).
         If None, will estimate based on
     method : str
-        Description of parameter `method` (the default is 'TSNE').
+        embedding method to use (the default is 'UMAP').
     perplexity : float
-        Description of parameter `perplexity` (the default is 30).
+        hyperparameter used to tune TSNE (the default is 30.0).
+        See below for details:
+            https://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html#sklearn.manifold.TSNE
```
jason-c-kwan: You could also reference our NAR paper, because there are a couple of paragraphs on the perplexity setting.

autometa/common/kmers.py
```diff
@@ -319,13 +430,13 @@ def embed(kmers=None, embedded=None, n_components=2, do_pca=True, pca_dimensions
...
-    scaler = 3.0
-    if n_rows < (scaler*perplexity):
-        perplexity = (n_rows/scaler) - 1
 
     def do_TSNE(perplexity=perplexity, n_components=n_components):
+        # Adjust perplexity according to the number of data points
+        n_rows = n_samples-1
+        scaler = 3.0
+        if n_rows < (scaler*perplexity):
+            perplexity = (n_rows/scaler) - 1
```
jason-c-kwan: There needs to be some documentation somewhere that this gives different results to the Python2.7 version. Perhaps in the docs or in comments somewhere.

autometa/common/coverage.py
jason-c-kwan: Yes, the same description is also at the top of the script in a comment.

autometa/binning/recursive_dbscan.py
WiscEvan: Ah, I see. I'll fix this comment.

autometa/common/kmers.py
WiscEvan: Going into detail here may be more appropriate for the readthedocs documentation. Although, when we have that up, we can reference that in the code. I'll put a COMBAK statement there for now

autometa/common/kmers.py
WiscEvan: I'll place a note, and we can place a link to a discussion on the differences in the readthedocs documentation.

autometa/common/external/prodigal.py
WiscEvan: I don't think there is a `validFasta` or some equivalent function in BioPython, but we can retrieve the records from the files with Biopython and if we do not encounter an error, assume they are valid.

autometa/common/kmers.py
WiscEvan: I've simplified the function to make it intuitive to the reader. 

I.e. complement the character for every character in the string in reverse order

I've also removed the unnecessary exception handling and denoted this a function used only as an internal by prepending the underscore (conventions to the programming community).

autometa/common/external/prodigal.py
jason-c-kwan: Yes, that is probably the next best thing, but note that it might not give an error with a zero size file (I guess that would technically be a valid fasta file with no contents). So perhaps check that `size != 0` and try loading with BioPython?

autometa/common/coverage.py
WiscEvan: I believe I've addressed the bowtie.py issues you raised? See [here](https://github.com/KwanLab/Autometa/blob/406041bcbca9e9e14fc3991cad8ad5104509de83/autometa/common/external/bowtie.py#L83-L137)

PR #29, opened 2020-03-18
Resolved #19 added docstring, fixed nproc and removed depth function

### LINKED ISSUE #19: [Dev branch] Issues with autometa/common/external/samtools.py
This file contains wrapper functions for samtools (so converting a sam alignment file to a sorted bam and calculating the read depth).

Issues:

1. In the `if __name__ == '__main__':` block, the argparse block doesn't contain any documentation as to what the program does or what the command line arguments are. 

2. The `main` function calls a function that is not implemented (`depth`, see below).

3. The `sort` function takes three arguments, `sam`, `out` and `nproc=0`, meaning that the default value for `nproc` is zero if it is not specified. I couldn't find anything in the samtools documentation about what happens when you specify zero processors, so it would be better to explicitly state this in a comment or perhaps change the default value. The function should probably also have a docstring.

4. The `depth` function is not implemented and it raises a NotImplementedError. It also does not have a docstring to explain, for example, what type of object `records` should be. (It looks like it could be a list of BioPython SeqRecords). Without a docstring it is unclear why this function is needed when we have the functions in bedtools.py.

5. The `run` function seems to handle the running of external programs in this script. It takes a string which is interpreted as a command by `subprocess.call`. The stdout and stderr are both directed to `os.devnull`. One glaring problem with this function are the return values. In Unix operating systems all programs return a value, and by convention it is zero if the program was successful and nonzero if not (the exact value will be specific to the type of error for that program). In programming languages, numbers can be cast as bools, with `0`, `None` or `undef` evaluating as `False` in Python (and many other languages), while nonzero evaluates as `True`. What this function does is exactly the opposite. If `retcode == 0` then it returns `True` and if `retcode != 0` it returns `False`. While this may seem intuitive, it is exactly the opposite of what most programmers will assume, and so I strongly advise that this is changed.

Sidduppal:
Issues resolved:
1. Added documentation in the argparse block, including help and type of input needed.
2. The `depth` function was not being called from any other script. The function is supposed to use samtools to compute the read depth at each position or region, however, it is not being implemented anywhere. It was removed after consulting with @WiscEvan.
3. The default value of `nproc` was changed to `mp.cpu_count()` it now uses all the processors in the system as a default option. 
4. Tried to make the `run` function more intuitive. Changes here might affect coverage.py file as the function is being called there.


autometa/common/external/samtools.py
```diff
@@ -61,8 +89,8 @@ def main(args):
         format='%(asctime)s : %(name)s : %(levelname)s : %(message)s',
         datefmt='%m/%d/%Y %I:%M:%S %p')
     parser = argparse.ArgumentParser()
-    parser.add_argument('--bam')
```
WiscEvan: I think `--out` instead of `--bam` is probably more appropriate. However, there should still be some documentation describing the output as a _bam_ file.

autometa/common/external/samtools.py
```diff
@@ -61,8 +89,8 @@ def main(args):
         format='%(asctime)s : %(name)s : %(levelname)s : %(message)s',
         datefmt='%m/%d/%Y %I:%M:%S %p')
     parser = argparse.ArgumentParser()
```
WiscEvan: You can place the description of the script when you initialize `argparse.ArgumentParser`. I.e. Perhaps this should be the place where you describe taking a sam file and converting it to a bam file

`parser = argparse.ArgumentParser(description='Description of script to display to user')`

autometa/common/external/samtools.py
```diff
@@ -61,8 +89,8 @@ def main(args):
...
-    parser.add_argument('--bam')
-    parser.add_argument('--sam')
-    parser.add_argument('--seqs')
+    parser.add_argument('--sam', help='</path/to/alignment.sam>')
+    parser.add_argument('--out', help='</path/to/output/file')
```
WiscEvan: Similarly, you could place the `.bam` extension on the output file like so, `</path/to/output/file.bam>`

autometa/common/external/samtools.py
```diff
@@ -37,22 +56,31 @@ def run(cmd):
...
+    """
+    Views the sam file and then sorts the alighnments by leftmost coordinates.
+    
+    Parameters
+    ----------
+    sam : str
+        </path/to/alignment.sam>
+    out : str
+        </path/to/output/file
+    nproc : int, optional
```
WiscEvan: Great catch on adding `optional`. I take it you read through the numpy [docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html)? This should probably be done throughout the code base. I was not aware of this previously. 💯

Sidduppal: Actually no, I just used VS code extension for docstring and just to be consistent with your code I used the numpy format. I can read through the guide now

WiscEvan: Well, in any case. The VS code formatted it appropriately. 👍

autometa/common/external/samtools.py
```diff
@@ -37,32 +56,42 @@ def run(cmd):
...
         logger.warning(f'args:{cmd} retcode:{retcode}')
         return False
-    return True
 
-def sort(sam, out, nproc=0):
-    cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {out}'
-    run(cmd)
+def sort(sam, bam, nproc=mp.cpu_count()):
+    """
+    Views the sam file and then sorts the alighnments by leftmost coordinates.
```
WiscEvan: 1. `alighnments` typo.
2. You can shorten the docstring by denoting the inputs in the docstring.
I.e. Views then sorts `sam` by leftmost coordinates.

autometa/common/external/samtools.py
```diff
@@ -37,32 +56,42 @@ def run(cmd):
...
+    Ouputs the sorted alighnments in the specified bam file.
```
WiscEvan: You could probably place this in the previous sentence.

I.e. Views then sorts `sam` by leftmost coordinates and outputs to `bam`

autometa/common/external/samtools.py
WiscEvan: You could similarly call the output `file.bam` an `alignment.bam` file. Although this is pretty minor and inconsequential. Up to you

autometa/common/external/samtools.py
```diff
@@ -37,32 +56,41 @@ def run(cmd):
     logger.debug(f'cmd: {cmd}')
     with open(os.devnull, 'w') as stdout, open(os.devnull, 'w') as stderr:
         retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
-    if retcode:
+    if not retcode:
+        return True
+    else:
         logger.warning(f'args:{cmd} retcode:{retcode}')
         return False
-    return True
```
WiscEvan: I think it is clearer using `if retcode` versus `if not retcode`

Similarly an else statement is, in this case, redundant because if the "if" statement is executed, the function will return a value. Jason's comments on the return values here are specific to programming conventions. Since the return values are not used, you could return a NoneType object and in the case of the `run` command failling, raise a ChildProcessError instead of a logger warning.

You could tackle this two ways:

1. If there is some issue, I.e. a return code that is not equal to 0, then raise a ChildProcessError (this is a python built-in) exception. Note: The docstring will need to be updated to reflect this Exception.
2. Log the exception via logger.exception('same string as the logger.warning string') and allow Autometa to continue the analysis. 

**My preference lies with number 1.**

autometa/common/external/samtools.py
```diff
@@ -26,58 +26,65 @@
...
 import pandas as pd
 import multiprocessing as mp
 
 logger = logging.getLogger(__name__)
 
 
-def run(cmd):
-    """Run `cmd` via subprocess.
-
+def run(cmd, bam):
```
WiscEvan: If you're incorporating \`bam\` into the run function, you could probably just incorporate `run` into `sort`.  Since no other functions exist in the script that would call run. This may simplify things further

autometa/common/external/samtools.py
```diff
@@ -26,58 +26,65 @@
...
+    ChildProcessError
+        Function did not run successfully, returns retcode.
     """
+    
     logger.debug(f'cmd: {cmd}')
-    with open(os.devnull, 'w') as stdout, open(os.devnull, 'w') as stderr:
+    log_samtools_dir = os.path.dirname(bam)
+    tempdir = tempfile.mkdtemp(suffix=None, prefix='samtools', dir=log_samtools_dir)
+    samtools_stderr=os.path.join(tempdir, 'samtools_stderr')
+    samtools_stdout=os.path.join(tempdir, 'samtools_stdout')
```
WiscEvan: Typically stderr and stdout are denoted as `.err` and `.out` respectively. If you like, you could also keep the stderr and stdout but add them as extensions to the samtools file name. I.e. `samtools.stderr` or `samtools.err` and `samtools.stdout` or `samtools.out`

autometa/common/external/samtools.py
```diff
@@ -26,58 +26,65 @@
...
+    with open(samtools_stdout, 'w') as stdout, open(samtools_stderr, 'w') as stderr:
         retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
-    if not retcode:
+        if retcode != 0:
+            raise ChildProcessError(f'Samtools not successfully run, retcode: {retcode}')
+        shutil.rmtree(tempdir, ignore_errors=True)
```
WiscEvan: Following lines 63-64 above, Perhaps it would be better to remove all intermediate files (whether successful or not) and keep the logs only if the process fails. If this step continues to fail, the alignment files could take up a lot of unnecessary space, where as the logs should not take up too much space.

autometa/common/external/samtools.py
```diff
@@ -81,11 +43,33 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+    ChildProcessError
+        Samtools did not run successfully, returns retcode.
+    
+    """
 
+    cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {bam}'
+    logger.debug(f'cmd: {cmd}')
+    samtools_out_dir = os.path.dirname(bam)
+    tempdir = tempfile.mkdtemp(suffix=None, prefix='samtools', dir=samtools_out_dir)
+    os.chdir(tempdir)
```
WiscEvan: Is changing to the samtools temp dir necessary here?

autometa/common/external/samtools.py
```diff
@@ -100,5 +84,4 @@ def main(args):
     parser.add_argument('--bam', help='</path/to/output/alignment.bam>', type=str)
     parser.add_argument('--nproc', help='Number of processors to use', default=mp.cpu_count(), type=int)
     args = parser.parse_args()
-    main(args)
-
+    main(args)
```
WiscEvan: It appears GitHub is telling us there should be a newline at the end of the file.

autometa/common/external/samtools.py
```diff
@@ -81,11 +43,33 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+    Raises
+    ------
+    ChildProcessError
+        Samtools did not run successfully, returns retcode.
+    
+    """
 
+    cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {bam}'
+    logger.debug(f'cmd: {cmd}')
+    samtools_out_dir = os.path.dirname(bam)
```
WiscEvan: If the \`bam\` provided is not a full path then the dirname could be empty. You could account for this with

`samtools_out_dir = os.path.dirname(os.path.abspath(bam))`

autometa/common/external/samtools.py
Sidduppal: I added a new line. Although is there any specific reason to do it? I mean what would happen if we don't. Just curious.

WiscEvan: See discussion here: https://stackoverflow.com/a/5813359

autometa/common/external/samtools.py
jason-c-kwan: Agree

autometa/common/external/samtools.py
jason-c-kwan: I am also unclear why you need the temp directory. If nothing else it is weird to change to that directory and then later to delete it. Not sure what the OS will do in that situation, and also other scripts might assume that the directory they are in doesn't change (although I guess it is good practice not to assume that sort of thing and use full paths where possible).

autometa/common/external/samtools.py
```diff
@@ -51,24 +51,29 @@ def sort(sam, bam, nproc=mp.cpu_count()):
     
     """
 
-    cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {bam}'
+    samtools_out_dir = os.path.dirname(os.path.abspath(bam))
+    outfile_tail = os.path.basename(bam)
```
WiscEvan: What is the reasoning for the tail? I'm a bit confused with the variable name here.

autometa/common/external/samtools.py
```diff
@@ -51,24 +51,29 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+    tempdir = tempfile.mkdtemp(suffix = None, prefix = 'samtools', dir = samtools_out_dir)
+    samtools_outfile_tail = os.path.join(samtools_out_dir, outfile_tail)
+    temp_outfile_tail = os.path.join(tempdir, outfile_tail)
+    cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {temp_outfile_tail}'
```
WiscEvan: There are currently no checks for the `nproc` parameter to be an integer greater than or equal to one. If None is provided, does this result in an error? 0?

autometa/common/external/samtools.py
```diff
@@ -51,24 +51,29 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
                 raise ChildProcessError(f'Samtools not successfully run, retcode: {retcode}')
+            else:
+                shutil.move(temp_outfile_tail, samtools_outfile_tail)
         except ChildProcessError as err:
             raise err     
         finally:
             shutil.rmtree(tempdir, ignore_errors=True)
     os.remove(samtools_err)
     os.remove(samtools_out)
+    logger.debug(f'Samtools run successfully ! bam output in {samtools_out_dir}')
```
WiscEvan: This logger message seems unnecessary or could you place this in the `main` function so the logger is not too verbose? We should be able to tell with the previous logging and the `shutil` and `os` calls whether samtools sort appropriately ran.

autometa/common/external/samtools.py
```diff
@@ -51,24 +51,29 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-    samtools_out_dir = os.path.dirname(bam)
-    tempdir = tempfile.mkdtemp(suffix=None, prefix='samtools', dir=samtools_out_dir)
-    os.chdir(tempdir)
     samtools_err = os.path.join(samtools_out_dir, 'samtools.err')
     samtools_out = os.path.join(samtools_out_dir, 'samtools.out')
     with open(samtools_out, 'w') as stdout, open(samtools_err, 'w') as stderr:
         try:
             retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
             if retcode != 0:
                 raise ChildProcessError(f'Samtools not successfully run, retcode: {retcode}')
```
WiscEvan: You could shorten this statement to say "sort failed" rather than "not successfully run". (Minor edit to help screen wrapping). This is more descriptive, to the function in which the child process is being raised and more concise. When the exception is raised, the traceback will need to the filename hence "samtools" will already be referenced.

autometa/common/external/samtools.py
```diff
@@ -51,24 +51,29 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+            else:
+                shutil.move(temp_outfile_tail, samtools_outfile_tail)
```
WiscEvan: Maybe it would be more clear if you kept the variables resembling the sam and bam filepaths to which they're pointing? I'm still confused as to where the tail portion of the variable name is necessary.

autometa/common/external/samtools.py
Sidduppal: It indicates the file name. Like, ``/home/siddharth/Downloads/align.bam`` tail was supposed to indicate ``align.bam``. I could replace ``tail`` everywhere with ``name`` is that more appropriate?

autometa/common/external/samtools.py
Sidduppal: The ``tail`` portion from ``temp_outfile_tail`` and ``samtools_outfile_tail`` has been removed.

autometa/common/external/samtools.py
```diff
@@ -46,34 +46,33 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-                shutil.move(temp_outfile_tail, samtools_outfile_tail)
-        except ChildProcessError as err:
-            raise err     
-        finally:
+        retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
+        if retcode != 0:
+            shutil.rmtree(tempdir, ignore_errors=True)
+            raise ChildProcessError(f'Sort failed, retcode: {retcode}')
+        else:
+            shutil.move(temp_outfile, samtools_outfile)
```
WiscEvan: This will work in according to the behavior we expect for the program, however, if a different exception is raised (for example the user kills the pipeline) will the temporary directory remain? We also need to account for unexpected behavior, which can be difficult due to the nature of the issue, although try/except/finally _and_ specific exception handling can help assess any of the known issues to provide _some_ clarity to the unknown issues.

You might be able to achieve the required behavior (unsure about the exception handling) strictly using the [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) function from the built-in tempfile library. 

- [tempfile examples](https://docs.python.org/3/library/tempfile.html#examples)
- [Unified try/except/finally](https://docs.python.org/2.5/whatsnew/pep-341.html)

Sidduppal: I can check for ``FileNotFoundError`` (the same way I am checking if ``nproc`` > 0) but I am not exactly sure what all other errors might we encounter and how can we catch-all of them individually.
Won't they be recorded in ``samtool.err`` file

autometa/common/external/samtools.py
```diff
@@ -47,32 +47,30 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+    with tempfile.TemporaryDirectory(suffix = None, prefix = 'samtools', dir = samtools_out_dir) as tempdir:    
+        temp_outfile = os.path.join(tempdir, outfile)
+        cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {temp_outfile}'
+        logger.debug(f'cmd: {cmd}')
+        with open(samtools_out, 'w') as stdout, open(samtools_err, 'w') as stderr:
+            retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
+            if retcode != 0:
+                raise ChildProcessError(f'Sort failed, retcode: {retcode}')
+            else:
+                shutil.move(temp_outfile, samtools_outfile)
```
WiscEvan: This does not need to be within the second with statement. The `if/else` clause here should be one indent less.

autometa/common/external/samtools.py
```diff
@@ -47,32 +47,30 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-            shutil.rmtree(tempdir, ignore_errors=True)
-            raise ChildProcessError(f'Sort failed, retcode: {retcode}')
-        else:
-            shutil.move(temp_outfile, samtools_outfile)
-            shutil.rmtree(tempdir, ignore_errors=True)
-    os.remove(samtools_err)
-    os.remove(samtools_out)
+    outfile = os.path.basename(bam)
+    samtools_outfile = os.path.join(samtools_out_dir, outfile)
+    with tempfile.TemporaryDirectory(suffix = None, prefix = 'samtools', dir = samtools_out_dir) as tempdir:
```
WiscEvan: A few things here

1. When passing arguments into a function, recommended style is to  place args with their keyword. I.e. `suffix=None, prefix='samtools', dir='samtools_out_dir'`
2. Since you are passing `None` into suffix, you can omit this.
3. Do any of these arguments need to be supplied? Do we care where the tmpdir is located and do we care what it is named? If no, we can omit all of these arguments and this should allow you to be more concise, allowing for better readability.

I have been following this styling guide: [google style guide - whitespace](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#36-whitespace)

autometa/common/external/samtools.py
```diff
@@ -47,32 +47,30 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-        if retcode != 0:
-            shutil.rmtree(tempdir, ignore_errors=True)
-            raise ChildProcessError(f'Sort failed, retcode: {retcode}')
-        else:
-            shutil.move(temp_outfile, samtools_outfile)
-            shutil.rmtree(tempdir, ignore_errors=True)
-    os.remove(samtools_err)
-    os.remove(samtools_out)
+    outfile = os.path.basename(bam)
+    samtools_outfile = os.path.join(samtools_out_dir, outfile)
```
WiscEvan: How is this different than the originally supplied \`bam\` argument?

This is redundant, you are taking the directory name initially and now you are joining the path that already exists.

```python
# original outfile path with same directory is already in the `bam` variable

samtools_out_dir =os.path.dirname(os.path.abspath(bam))

outfile = os.path.basename(bam)
samtools_outfile = os.path.join(samtools_out_dir, outfile)

# These should be equal? So the redundant code should be removed.
samtools_outfile == bam
```

autometa/common/external/samtools.py
```diff
@@ -47,32 +47,30 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
         raise TypeError(f'nproc must be an integer greater than zero! Given: {nproc}')
     samtools_out_dir = os.path.dirname(os.path.abspath(bam))
-    outfile = os.path.basename(bam)
-    tempdir = tempfile.mkdtemp(suffix = None, prefix = 'samtools', dir = samtools_out_dir)
-    samtools_outfile = os.path.join(samtools_out_dir, outfile)
-    temp_outfile = os.path.join(tempdir, outfile)
-    cmd = f'samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {temp_outfile}'
-    logger.debug(f'cmd: {cmd}')
     samtools_err = os.path.join(samtools_out_dir, 'samtools.err')
     samtools_out = os.path.join(samtools_out_dir, 'samtools.out')
```
WiscEvan: This is a minor nitpick, but you could just name these variables `err` and `out`, because the entire script is within the context of samtools. This may improve readability.

```python
err = os.path.join(samtools_out_dir, 'samtools.err')
out = os.path.join(samtools_out_dir, 'samtools.out')
with open(out, 'w') as stdout, open(err, 'w') as stderr:
    ...
```

autometa/common/external/samtools.py
```diff
@@ -47,32 +47,30 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-            raise ChildProcessError(f'Sort failed, retcode: {retcode}')
-        else:
-            shutil.move(temp_outfile, samtools_outfile)
-            shutil.rmtree(tempdir, ignore_errors=True)
-    os.remove(samtools_err)
-    os.remove(samtools_out)
+    outfile = os.path.basename(bam)
+    samtools_outfile = os.path.join(samtools_out_dir, outfile)
+    with tempfile.TemporaryDirectory(suffix = None, prefix = 'samtools', dir = samtools_out_dir) as tempdir:    
+        temp_outfile = os.path.join(tempdir, outfile)
```
WiscEvan: It may be more clear to just pull the basename into the `tmpdir` context and then you only need to move the `tmp_bam` to `bam`


`tmp_bam = os.path.join(tempdir, os.path.basename(bam))`

Then

`shutil.move(tmp_bam, bam)`

autometa/common/external/samtools.py
Sidduppal: I think yes in most cases, but if someone provides a relative or incomplete path then probably no. I did taking your previous your review in mind: https://github.com/KwanLab/Autometa/pull/29#discussion_r396533403.

WiscEvan: The comment you've linked is in reference to extracting the directory from the `bam`, in which case you would need to account for relative and absolute paths. However, the directory should be inconsequential, so you will not have to account for this.

WiscEvan: Notice, `samtools_out_dir` will never have to be utilized, therefore determining the directory of the provided path is unnecessary.

autometa/common/external/samtools.py
Sidduppal: I don't think we care where it is located and what it is named. I looked at some of the scripts where you have made a ``temp`` directory and since you had supplied these arguments as well I thought it is a good practice to supply them.

autometa/common/external/samtools.py
Sidduppal: But in case of some error, if the file is labeled ``samtools.err`` or ``samtools.out`` won't it help the user in identification?

WiscEvan: The logger will tell the user where the error occurred and the samtools files that were written may provide insight as to what went wrong in samtools

PR #35, opened 2020-04-02
nr.gz database downloading bug

WiscEvan:
- Found an issue with `check_dbs` where autometa will loook for `nr.dmnd.md5` and if it is not found, will update nr. However, when nr is updated, the nr.dmnd.md5 is never written. 
- Fixed file path handling to be less arch specific 
- updated variables for readability.

### DIFF
--- a/pipeline/make_taxonomy_table.py
+++ b/pipeline/make_taxonomy_table.py
@@ -27,7 +27,12 @@ import pandas as pd
 from argparse import ArgumentParser
 from Bio import SeqIO
 
-def run_command(command_string, stdout_path = None):
+
+PIPELINE = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


WiscEvan:
This has not been performance tested outside of running the script to see if any error messages were emitted.

jason-c-kwan:
Do you at least know if the error is corrected? From looking through the code it seems to look for nr.gz.md5 now instead of nr.dmnd.md5. Is that correct?

WiscEvan:
There exists an nr.gz checksum on NCBI's FTP server (See here [nr.gz.md5](https://ftp.ncbi.nlm.nih.gov/blast/db/FASTA/nr.gz.md5) Where the diamond formatted database checksum is not hosted anywhere.

jason-c-kwan:
OK, well after I merge, please check with the user that this fixes their problem.

PR #37, opened 2020-04-06
Added missing merged.dmp in database checking

WiscEvan:
Added missing merged.dmp in database checking when using default autometa databases directory

### DIFF
--- a/pipeline/make_taxonomy_table.py
+++ b/pipeline/make_taxonomy_table.py
@@ -189,7 +189,7 @@ def update_dbs(outdir, db='all'):
 		if os.path.isfile(taxdump_fpath):
 			run_command('tar -xzf {} -C {} names.dmp nodes.dmp merged.dmp'.format(taxdump_fpath, outdir))
 			os.remove(taxdump_fpath)
-			print("nodes.dmp,names.dmp, merged.dmp updated")
+			print("nodes.dmp, names.dmp, merged.dmp updated")
 
 def check_dbs(db_path):
 	'''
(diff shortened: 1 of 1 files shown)


WiscEvan:
I agree. checksums are probably our best bet. Some functionality for this has already been implemented in dev.

jason-c-kwan:
Perhaps there could be a file class that is responsible for keeping track of whether a file was created by the program (in which case a SH1 sum is made upon completion), or downloaded (in which case it looks for the md5 at the FTP site, and if that doesn't exist it makes one. This might be a stretch goal for right now though.

WiscEvan:
There are some functions already implemented and available in the `dev` branch:

## `get_checksum()`

https://github.com/KwanLab/Autometa/blob/59941204e1ebb59a04da01b9bbd46c7fd50d3e7b/autometa/common/utilities.py#L256-L299

## `get_checksum()` applied in checkpoint functions:

https://github.com/KwanLab/Autometa/blob/59941204e1ebb59a04da01b9bbd46c7fd50d3e7b/autometa/common/utilities.py#L301-L411

## Notes on current ideas about (tentative name) "Checkpoints" class layout


The functions above should be be added into the class or removed once the class is implemented.

I have thought about a class to handle this functionality. I think this would be quite useful when constructing a workflow with [Work Queue](http://ccl.cse.nd.edu/software/workqueue/).

Here are my current notes on a `Checkpoints` class:

- init string requires project directory as input. (i.e. `Checkpoints(</path/to/project/directory>)`)
    - Directory and subdirectories will be recursively searched for all `checkpoints.txt` files.
    - This allows query of progress and data state for any metagenome of interest in the project directory.
- Method should make setting a checkpoint and retrieving a checkpoint as easy as possible.
    - This could be something like a decorator to allow `@checkpoint` at a function definition with the input and output files (In this case I think this would have to be a staticmethod of the class)
- One method (which would be useful to integrate with work-queue) to get the status of a metagenome given the input metagenome as the arg outputting the annotations still required.
- A method to clear the progress of a metagenome directory.
- A method to clear all progress for a whole project directory.

It may be most appropriate to open another issue specifying to build a `Checkpoints` class with the desired functionality as this issue is a _hotfix_ for one database file.

Issue #5, opened 2020-02-06
create autometa conda recipe


WiscEvan:
Write a conda recipe to allow users to install autometa and its dependencies with the simple command:

 `conda install -c bioconda autometa`

Note, this command assumes the autometa conda recipe is housed in the [_bioconda_](https://bioconda.github.io/) channel.

Conda recipes tutorial - [link](https://python-packaging-tutorial.readthedocs.io/en/latest/conda.html "python-packaging-tutorial")

closed Autometa-2.0.0a PR - https://github.com/bioconda/bioconda-recipes/pull/21480

chasemc:
nf-core linter expects a `environment.yml`. While we can turn off that lint check I think it may be beneficial to consider switching from `requirements.txt` to `environment.yml`
![image](https://user-images.githubusercontent.com/18691127/122279212-6b4ffb80-cead-11eb-80bb-1f05d77e1029.png)

WiscEvan:
Hosted https://anaconda.org/bioconda/autometa

PR #38, opened 2020-04-08
Issue #5 Working conda recipe

WiscEvan:
# Autometa: User friendly installation & usage

Managed to figure out the conda recipe for _so far_ the easiest install available.

**Full installation:** `conda install -c erees autometa`

For reference see anaconda cloud [erees autometa package](https://anaconda.org/erees/autometa)

When this is merged into dev we can integrate with the `bioconda` channel. This will change the installation to `conda install -c bioconda autometa`, similar to many other bioinformatics tools (such as hmmer, spades, bowtie2... the list is in the thousands). 

Another convenience of hosting the package on the `bioconda` channel is that many bioinformatics people will often add `bioconda` as one of their default channels in their conda config. Ultimately, with `conda config --append channels bioconda`: The one-liner becomes: `conda install autometa`

The conda recipe allows continuous integration into these builds with bioconda so that with each new release a new build will be run and automatically hosted for users to update.

They will then be able to update via `conda update -n <their env name> autometa` and new user will have the most recent code changes.

## A few notes for double-checking and consideration:

Autometa now operates with console scripts (AKA entry points). These require the _main_ function found within each respective module to be _empty of any arguments._ Therefore there are a few structural changes to the template as well as other modules found within the autometa code base that we would like to expose to the common end user. The structural changes to the code are reflected in `docs/template.py` and of the _41_ files changes, many of these changes are just altering this structure to create our console scripts.

Other major changes to the structure include moving the `databases` directory _within_ the autometa _package_. This is necessary for automatically handling database dependencies strictly within the autometa package and not somewhere else on the user's environment. When packaging typically only scripts are added to reduce the package size. Therefore, directories strictly to fill with data (databases) are omitted. The `autometa-configure` console script, run immediately after install via `conda install autometa` will download and format all required databases into the default autometa directory. Similarly, dependencies can be checked with `autometa` _and_ `autometa-configure`.

...On to the lower level dev questions/suggestions.

### Console scripts now available to the end user:

- [x] autometa
- [x] autometa-configure
- [x] autometa-kmers
- [x] autometa-coverage
- [x] autometa-markers
- [ ] autometa-taxa
- [ ] autometa-binning

The checked boxes above represent the currently available console scripts. I think it would be useful to the end user to expose a taxon assignment entry point. However, I do not believe we currently have one outside of calling `majority_vote.py`. I think this may require opening another issue to have a general script that will incorporate any method of taxon assignment we add or is currently in use. Taxon assignment is currently imported and used within the `metagenome.py` script from `majority_vote.py`, but perhaps an import from taxon_assignment would be more appropriate for modularity (i.e. easier maintenance, tests, CI/CD, etc.).

As I was performing this restructuring another question arose on whether to alter the `workspace` parameter to either `outdir` or to remove this entirely.
The `workspace` is constructed to contain all of the user's projects, although I'm not entirely sure it is necessary.  I initially wrote this with the website in mind, but maybe @chanana can provide some details on whether or not this can be removed. This would change the input config file a bit where the `project_num` parameter could simply be `outdir` and the `metagenome` parameters can each correspond to their respective metagenome that will be housed within the `outdir` which would serve as the _project_. I imagine Django could construct its own user-specific file system hierarchy which will not require this `workspace` structure. Therefore to allow more fine-grained control of the output directory for any users who would only like to use the CLI we could go this route.

**I've also deleted the jupyter notebooks in the `docs` directory as these are out of date compared to the updated documentation**

## conda builds

The conda recipe is housed within the bioconda repo. Bioconda has a bot that will trigger whenever (see [build system](http://bioconda.github.io/contributor/build-system.html#build-system)) a pull request or any changes are made. I think that is it. If this is specific to bioconda, then I believe we can set up a GitHub action that will handle updating using Bioconda's [updating recipes documentation](http://bioconda.github.io/contributor/updating.html#updating-recipes).

Unfortunately I did not find a "one size fits all" tutorial that worked perfectly. Had to amass some insights from different sources to successfully reach the completed build.

Feel free to test this out yourself by checking out my branch...

**Note: since this is a build, best to have just the repo without any of your untracked files, this may result in unexpected behavior.** 

... So you should clone a clean Autometa repo to try this out...

```bash
# You will need conda-build installed.
conda install -n <your build env name> conda-build conda-verify

# Grab the repo and check out this PR.
cd ~/ && git clone github.com/KwanLab/Autometa.git && cd ~/Autometa
git remote add WiscEvan github.com/WiscEvan/Autometa.git
git fetch WiscEvan
git checkout -b evan-conda WiscEvan/issue-#5

# Try the build.
conda activate <your build env name>
conda build .
```

## References:

- Conda build process from conda - [link](https://docs.conda.io/projects/conda-build/en/latest/concepts/recipe.html)
- bioconda contribution workflow - [link](http://bioconda.github.io/contributor/workflow.html#contribution-workflow)
- writing the setup script - [link](https://docs.python.org/3/distutils/setupscript.html)
- setuptools documentation - [link](https://setuptools.readthedocs.io/en/latest/)
- entry points - [link](https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/)


jason-c-kwan:
After this is merged, would the conda recipe just pull from the releases? (and ultimately from master?)

jason-c-kwan:
Also, could you post a link to the best information about how to make conda recipes? I think that will help me review.

WiscEvan:
The conda recipe is housed within the bioconda repo. Bioconda has a bot that will trigger whenever (see [build system](http://bioconda.github.io/contributor/build-system.html#build-system)) a pull request or any changes are made. I think that is it. If this is specific to bioconda, then I believe we can set up a GitHub action that will handle updating using Bioconda's [updating recipes documentation](http://bioconda.github.io/contributor/updating.html#updating-recipes).

## References:

- Conda build process from conda - [link](https://docs.conda.io/projects/conda-build/en/latest/concepts/recipe.html)
- bioconda contribution workflow - [link](http://bioconda.github.io/contributor/workflow.html#contribution-workflow)

jason-c-kwan:
Another question - does this mean that the user no longer has the option of having databases somewhere other than the Autometa directory?

autometa/common/external/work_queue.py
```diff
@@ -144,7 +144,7 @@ def main(args):
     q.activate_fast_abort(2)
 
     # create task as usual and tag it with an arbitrary string.
-    t = wq.Task(...)
+    # t = wq.Task(...)
```
jason-c-kwan, line 147: So I haven't gone through this script yet. Could you perhaps comment on what is going on here, and the reason for this change?

VERSION
```diff
@@ -0,0 +1 @@
+2.0.0
```
jason-c-kwan, line 1: OK, so is it good practice to have this hard coded in while we are developing? I know we are working towards this. Perhaps it should be something like 2.0.0beta1? Or even alpha1? I'm just thinking of the version guidelines shown here: https://setuptools.readthedocs.io/en/latest/setuptools.html. Ultimately if any of these pre-release versions get into a repository somewhere they should be listed behind the ultimate 2.0.0.

setup.py
```diff
@@ -0,0 +1,46 @@
...
+        'console_scripts':[
+            'autometa = autometa.main:entrypoint',
+            'autometa-configure = autometa.config.user:main',
+            'autometa-kmers = autometa.common.kmers:main',
+            'autometa-coverage = autometa.common.coverage:main',
+            'autometa-markers = autometa.common.markers:main',
+        ]
+    },
+    author='Jason C. Kwan',
+    author_email='jkwan@wisc.edu',
```
jason-c-kwan, line 31: Can you change the email to jason.kwan@wisc.edu

setup.py
```diff
@@ -0,0 +1,46 @@
...
+    description='Automated Extraction of Genomes from Shotgun Metagenomes',
+    long_description=long_description,
+    long_description_content_type='text/markdown',
+    url='https://github.com/WiscEvan/Autometa',
```
jason-c-kwan, line 35: Remember to change this link to the main repo before merging

setup.py
```diff
@@ -0,0 +1,46 @@
...
+    license='GNU Affero General Public License v3 or later (AGPLv3+)',
+    classifiers=[
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 3',
+        'Development Status :: 3 - Alpha',
```
jason-c-kwan, line 40: This should be consistent with the version above?

meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+{% set data = load_setup_py_data() %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ data.get('version') }}
+
+source:
+  path: ./
+  # git_url: https://github.com/WiscEvan/Autometa.git
+  # git_rev: 2.0.0-cbld
```
jason-c-kwan, line 11: See other comments - link should be updated to main repo and version number should match to VERSION.

meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  # git_depth: 1
+
+build:
+  noarch: python
+  script: python -m pip install . --ignore-installed --no-deps -vv
+
+requirements:
+  host:
+    - python >=3
```
jason-c-kwan, line 20: Does not match the python >=3.7 requirement in `setup.py`.

meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - pip
+    - setuptools
+  run:
+    - python >=3
```
jason-c-kwan, line 24: See above

meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - autometa
+  commands:
+    - autometa -h
+    - autometa-configure -h
+    - autometa-kmers -h
+    - autometa-coverage -h
+    - autometa-markers -h
+
+about:
+  home: https://github.com/WiscEvan/Autometa
```
jason-c-kwan, line 54: Should update this link to the main repo

.condarc.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - defaults
+  - bioconda
+  - conda-forge
```
jason-c-kwan, line 4: These are the channels you need during setup, right?

jason-c-kwan:
A general comment - it would be good to have a fuller description of Autometa on the anaconda/bioconda page for this package. @WiscEvan your conda page doesn't have much information at present.

autometa/main.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python
```
jason-c-kwan, line 1: In the link you provided they used `__main__.py` and here you use `main.py`. Is this a convention that must be followed?

autometa/main.py
```diff
@@ -0,0 +1,159 @@
...
+
+import multiprocessing as mp
+
+from .config.user import AutometaUser
+
+
+logger = logging.getLogger('autometa')
+
+
+__version__ = "2.0.0"
```
jason-c-kwan, line 40: Make sure the versions match (see other comments)

jason-c-kwan: Also, perhaps should just reference `VERSION` so that you don’t have to alter the version in so many places each time.

autometa/common/external/work_queue.py
WiscEvan: This script is not functional right now. It will start up a workqueue master then raise a NotImplementedError

VERSION
WiscEvan: Yeah, sounds good.

.condarc.yaml
WiscEvan: Yes, conda-forge and bioconda channels host the packages autometa requires

autometa/main.py
WiscEvan: I'll remove this from main

autometa/main.py
WiscEvan: I can change this, I'm not familiar enough with packaging/distribution conventions to know one way or the other.

VERSION
WiscEvan: From [here](https://gamedev.stackexchange.com/a/75814) and [here](https://blog.codinghorror.com/alpha-beta-and-sometimes-gamma/) I believe we fit into the pre-alpha development status.

meta.yaml
WiscEvan: I am applying a jinja template to pull the version from `setup.py`. Although pulling from git has been commented out and is not being used in the build process.

`git_rev = {{ data.get('version') }}`

WiscEvan:
From https://github.com/KwanLab/Autometa/pull/38#issuecomment-611531811, the summary can be updated on the anaconda/bioconda page and similarly the summary can be written when the upload is performed. I'll double check if this can be pulled from `setup.py` or `meta.yaml`

autometa/main.py
WiscEvan: I've also templated all of the instances of version for meta.yaml from the one instance within `setup.py`. Should only have to change `VERSION` now and everything will be updated appropriately.

VERSION
WiscEvan: Defined to `2.0a0` based on this https://www.python.org/dev/peps/pep-0440/#id34

WiscEvan:
From https://github.com/KwanLab/Autometa/pull/38#issuecomment-611484596:

The user can pass a different directory to their `default.config` . I could also add an option to argparse to update given a provided directory path.

autometa.py
```diff
@@ -1,153 +1,12 @@
...
-(at your option) any later version.
 
-Autometa is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-GNU Affero General Public License for more details.
 
-You should have received a copy of the GNU Affero General Public License
-along with Autometa. If not, see <http://www.gnu.org/licenses/>.
-COPYRIGHT
```
jason-c-kwan, line 22: The copyright should still be here, shouldn't it?

autometa/__main__.py
```diff
@@ -0,0 +1,157 @@
...
+        # user.process_binning()
+    # user.get_pangenomes()
+
+def entrypoint():
+    import argparse
+    import time
+    cpus = mp.cpu_count()
+    parser = argparse.ArgumentParser(description='Main script to run Autometa pipeline.')
+    parser.add_argument('config',
+        help='</path/to/metagenome.config>',
```
jason-c-kwan, line 124: As this is a user-facing help message, this should be more friendly. For example "Path to your config file". Also - if this is a positional argument, how do you specify more than one config file?

autometa/__main__.py
```diff
@@ -0,0 +1,157 @@
...
+        nargs='*')
+    parser.add_argument('--cpus',
+        help=f'Num. cpus to use when updating/constructing databases (default: {cpus} cpus)',
+        type=int,
+        default=cpus)
+    parser.add_argument('--debug',
+        help='Stream debugging information to terminal',
+        action='store_true',
+        default=False)
+    parser.add_argument('--log', help='</path/to/autometa.log>', type=str)
```
jason-c-kwan, line 134: As this is a user-facing help message, you should make this more friendly. For example "Path to write a log file". Question that arises from this - should this be set from the config file? It might be nice to have that there as it would help people organize their outputs.

autometa.py
WiscEvan: I'll add this in

autometa/config/user.py
```diff
@@ -295,17 +302,39 @@ def run_binning(self, mgargs):
...
+def main():
     import argparse
+    import multiprocessing as mp
     import logging as logger
+    parser = argparse.ArgumentParser(description="""
+    Configures the Autometa user environment/databases.
+    Running without args will download and format Autometa database dependencies.
+    """)
+    parser.add_argument('--config',
+        help=f'</path/to/user.config> (default is {config.DEFAULT_FPATH}).',
```
jason-c-kwan, line 314: As this is a user-facing help message, should be more friendly. For example "Path to an Autometa .config file". Question - can more than one be specified? If so, should state explicitly in the help message so that people know.

autometa/config/user.py
```diff
@@ -295,17 +302,39 @@ def run_binning(self, mgargs):
...
+    parser.add_argument('--dryrun',
+        help='Log configuration without performing updates.',
+        action='store_true',
+        default=False)
+    parser.add_argument('--debug',
+        help='Stream debugging information to terminal.',
+        action='store_true',
+        default=False)
+    parser.add_argument('--cpus',
+        help=f'</path/to/user.config> (default is {mp.cpu_count()}).',
```
jason-c-kwan, line 325: Wrong help message here - this is not a path but an int. Should be something like "<number of threads to use>"

autometa/config/user.py
WiscEvan: No, this will be the base config. Separate from an input metagenome form

autometa/common/kmers.py
```diff
@@ -465,7 +465,34 @@ def do_UMAP(n_neighbors=15, n_components=n_components, metric='euclidean'):
...
+    import logging as logger
+    logger.basicConfig(
+        format='[%(asctime)s %(levelname)s] %(name)s: %(message)s',
+        datefmt='%m/%d/%Y %I:%M:%S %p',
+        level=logger.DEBUG)
+    skip_desc = '(will skip if file exists)'
+    cpus = mp.cpu_count()
+    parser = argparse.ArgumentParser(description='Count k-mer frequencies of given `fasta`')
+    parser.add_argument('--fasta', help='</path/to/sequences.fna>', required=True)
+    parser.add_argument('--kmers', help=f'</path/to/output/kmers.tsv> {skip_desc}', required=True)
```
jason-c-kwan, line 479: These help messages should tell the user what types of files these should be - for example "Metagenomic assembly fasta file", "K-mers tab-delimited frequency table (will create if it doesn't exist yet)".

autometa/common/kmers.py
```diff
@@ -465,7 +465,34 @@ def do_UMAP(n_neighbors=15, n_components=n_components, metric='euclidean'):
...
+    parser.add_argument('--size', help='k-mer size', default=5, type=int)
```
jason-c-kwan, line 480: Should probably say "k-mer size in bp"

autometa/common/kmers.py
```diff
@@ -465,7 +465,34 @@ def do_UMAP(n_neighbors=15, n_components=n_components, metric='euclidean'):
...
+    parser.add_argument('--normalized', help=f'</path/to/output/kmers.normalized.tsv> {skip_desc}')
+    parser.add_argument('--embedded', help='</path/to/kmers.embedded.tsv>')
```
jason-c-kwan, line 482: See other comment - these lines should tell the user what the file it is expecting is and whether the program will create it if it doesn't yet exist. We will be getting lots of questions - how do I make the embedded tsv?

autometa/common/kmers.py
```diff
@@ -465,7 +465,34 @@ def do_UMAP(n_neighbors=15, n_components=n_components, metric='euclidean'):
...
+    parser.add_argument('--method', help='embedding method', choices=['TSNE','UMAP'], default='UMAP')
+    parser.add_argument('--n-components', help='<num components of lower dimension manifold>',
+        type=int, default=2)
```
jason-c-kwan, line 485: This is a bit too jargonny for users. Suggest "Number of dimensions to reduce k-mer frequencies to"

autometa/common/kmers.py
```diff
@@ -465,7 +465,34 @@ def do_UMAP(n_neighbors=15, n_components=n_components, metric='euclidean'):
...
+    parser.add_argument('--do-pca', help='Whether to perform PCA prior to manifold learning',
+        action='store_true', default=False)
```
jason-c-kwan, line 487: Rather than manifold learning, how about "dimension reduction"? Makes more intuitive sense.

autometa/common/coverage.py
```diff
@@ -239,4 +219,24 @@ def main(args):
         default=False)
     parser.add_argument('--out', help='</path/to/coverages.tsv>', required=True)
```
jason-c-kwan, line 220: For this and other help messages (I couldn't get github to select the grey parts), need to explain more simply what the files are that the script is expecting. So rather than </path/to/coverages.tsv>, write "Path to write a table of coverages".

autometa/common/markers.py
```diff
@@ -201,21 +201,22 @@ def get(self, format='wide', **kwargs):
...
+def main():
     import argparse
     import logging as logger
     logger.basicConfig(
         format='%(asctime)s : %(name)s : %(levelname)s : %(message)s',
         datefmt='%m/%d/%Y %I:%M:%S %p',
         level=logger.DEBUG)
-    parser = argparse.ArgumentParser('')
+    parser = argparse.ArgumentParser(description='Annotate ORFs with kingdom-marker information')
     parser.add_argument('orfs', help='</path/to/prot.orfs.faa>')
```
jason-c-kwan, line 212: As this script may be run standalone by the user, the help messages have to be very clear. Here, you should explain what type of file you want, so "Path to a fasta file containing amino acid sequences of open reading frames".

autometa/common/markers.py
```diff
@@ -201,21 +201,22 @@ def get(self, format='wide', **kwargs):
...
     parser.add_argument('kingdom', help='kingdom to search for markers',
         choices=['bacteria','archaea'], default='bacteria')
-    parser.add_argument('--dbdir', help='</path/to/markers/dir>', default=MARKERS_DIR)
+    parser.add_argument('--dbdir', help=f'</path/to/markers/dir> {MARKERS_DIR}',
+        default=MARKERS_DIR)
```
jason-c-kwan, line 216: As the user might run this script standalone, the help messages have to be very clear. Here, it should be made clear what we mean. For example "Path to directory containing the single-copy marker HMM databases".

autometa/config/__init__.py
```diff
@@ -33,7 +33,6 @@
```
jason-c-kwan, line 33: I don't know if it is important, but this file does not have `COPYRIGHT` on either side of the copyright notice.

jason-c-kwan: Also, I'm not sure if a general docstring is needed here to explain the contents of the file.

autometa/__main__.py
WiscEvan: You can pass in multiple just separate your inputs with a space and argparse collects this into a list.

autometa/__main__.py
WiscEvan: We could place this as an additional file in metagenome.config under the `[files]` section

WiscEvan: This is outside of the scope of this issue and if we think this is necessary, we should add this as a separate issue.

autometa/config/user.py
WiscEvan: The `nargs='*'` argument in `add_argument` will tell argparse to collect multiple inputs. The reasoning behind `nargs='*'` (0 or more required) rather than `nargs='+'` (1 or more required) is due to the user having the option of not passing any metagenome.config file because they want to check their dependencies.

WiscEvan:
https://github.com/KwanLab/Autometa/pull/38#issuecomment-611651259

On second thought, the user should make edits in one part of the configuration. I.e. the default.config file. So adding this into the command line may be confusing. Going to leave changing any of the paths to be handled _only_ in the default.config file.

autometa/__main__.py
jason-c-kwan: Here we still have `help='</path/to/autometa.log>'`, when I think it would be friendlier to write something like "Path to write a log file".

Issue #54, opened 2020-04-30
[dev branch] Issues with autometa/common/metagenome.py


jason-c-kwan:
This file contains the definition of the `Metagenome` class and accompanying methods for handling data. In reviewing it, I noticed the following issues:

1. In the `main` function, the `--cpus` argument should be specified as an `int`. 

2. In the `Metagenome` docstring, it would be useful to specify that the attribute `size` is the size of the assembly in bp.

3. In the `__init__` method of the `Metagenome` class, a path to a taxonomy table must be supplied. During the normal course of business, this will be generated by other class methods. However, if `self.taxonomy_assigned` is `True`, then we load an existing file into a pandas DataFrame and assign it to `self.taxonomy`. The trouble is, it is unclear to me how `self.taxonomy_assigned` would ever be set to `True` when a `Metagenome` object is created, so this mechanism needs to be refined.

4. The `Metagenome` class has a number of methods with the `@property` decorator, which means they are like attributes which are calculated each time. The trouble is that a number of them rely on the `self.sequences` property. What this does is that it will parse through the assembly fasta file, and return a list of BioPython SeqIO objects. With a large assembly, this itself takes some time, and `self.sequences` is run for `self.nseqs`, `self.mean_gc`, `self.size`, `self.largest_seq`, `self.fragmentation_metric` (twice), `self.describe` (seven times!), and `length_filter`. I think either `self.sequences` should just load everything into memory, or perhaps if that is deemed to use too much memory, the method could populate all the dependent properties upon first running, and the dependent properties could instead be marked with `@cached_property`. This is a decorator defined in the `functools` module, which makes the method remember the return value after the first time. Connected with this issue, I am unclear on the difference between `self.sequences` and `self.seqrecords`, and the need for both methods.

5. In the `Metagenome` class, it might be better to have the `mean_gc` property calculate a length-weighted mean GC, because otherwise short contigs will skew the result because they invariably outnumber long contigs, and are less accurate in terms of GC. Perhaps a simple way of calculating this is just to total up the G + C (as well as the appropriate fraction of ambiguous bases), and express as a percentage of the total size.

6. The `orfs_called` property in the `Metagenome` class works out if `self.prot_orfs_fpath` and `self.nucl_orfs_fpath` both exist and are non-zero size. This is not a very sophisticated way of determining if the ORFs have been called because there could have been an error half way through the prodigal run, for instance. See also the `taxonomy_assigned` property.

[...]

Issue #53, opened 2020-04-28
[dev branch] Issues with autometa/common/mag.py


jason-c-kwan:
This file contains the definition of the MAG class, which is basically a subset of the contigs in a metagenome. The issues that I came across are outlined below:

1. [x] The name of the script and the class is perhaps not completely logical. Most of the time a MAG class represents a kingdom bin and not a metagenome-assembled-genome (MAG). There is also a problem with the name "bin", because this is already a function in Python that converts an integer to a binary string. Perhaps something like `meta_bin` would be more appropriate?

2. [x] The MAG class itself does not have a docstring.

3. [x] There are two attributes in the MAG class that are kind of similar - `self.assembly_name`, which is defined as `self.basename.split('.')[0]` and `self.root` which is defined as `os.path.splitext(self.basename)[0]`. It looks like `self.root` is used for `self.nucls_fname` and `self.prots_fname`, but is `self.assembly_name` used or needed?

4. [x] A number of the methods in the MAG class have the `@property` decorator, and so they are sort of like attributes but are computed when they are called upon. The problem here is that a number of them call `self.get_seqs`, which returns the BioPython SeqIO records for the MAG. This means that when `totalsize`, `size_pct`, `nallseqs`, `seqs_pct`, `size`, or `gc_content` are called, the fasta file the MAG is based on will be parsed each time. I would suggest the following - perhaps the `get_seqs` method could populate all the other values when it is first run, and this other things like `totalsize` could be converted from a `@property` to a `@cached_property` as defined in the [functools](https://docs.python.org/dev/library/functools.html) module. A `@cached_property` will remember the value after it is first run to save on re-computing time.

5. [x] The `gc_content` property in the MAG class uses a simple `np.mean` statement. I think it would be more appropriate to calculate a length-weighted mean of GC, because otherwise the GC value will tend to be skewed by short contigs, especially if the bin is made up of a lot of short contigs. The thing I am worried about here is that the GC content of short contigs is less likely to be completely representative of their parent genomes.

6. [x] Many of the methods and properties of the MAG class do not have docstrings, which is important now that we are using docstrings directly for the documentation. 

7. [ ] The `prepared` method seems to check whether the path given in the argument `fpath` exists and is non-zero size. I don't think this is a great way to determine if a step finished because the file could be there, be non-zero size, but still be corrupted.

[...]

Issue #33, opened 2020-03-31
[Dev branch] Issues with autometa/taxonomy/ncbi.py


jason-c-kwan:
This file contains the logic for handling the NCBI taxonomy databases. Presumably an equivalent would have to be made if we add other alternate taxonomies, like GTDB. I found the following issues in this file:

1. [x] It would be useful if the script's docstring stated that the file contains the definition of the `NCBI` class. At present it just says "Utilities file containing functions useful for handling NCBI taxonomy databases".

2. [x] At the top of the script a few constants are defined, `BASE_DIR`, which is the folder that contains the autometa folder, and `NCBI_DIR` which is defined as `BASE_DIR/databases/ncbi`. I just wanted to check if this is intended? I think I saw somewhere else that the default database directory is supposed to be INSIDE the autometa folder, rather than at the same level?

3. [ ] ~~In the `if __name__ == '__main__':` block, there is a description of the `--query-taxid` argument, but I don't think it adequately explains to the user what this is or what it does.~~

4. [x] In the `__init__` method, I am wondering if the `self.disable` variable is needed or ever used? It seems to be defined as `not self.verbose`, meaning that if `self.verbose = True` then `self.disable = False`. 

5. [x] In the `name` method, the docstring does not adequately explain what the `rank` argument is used for. From looking at the code, it seems that one can choose the rank of the name you want returned, so if you run `name(562, 'genus')` you would get back `'Escherichia'`, where 562 is the taxid for *Escherichia coli*.

6. [x] In several places the script attempts to check whether a taxid argument is an int. An example is shown below, but similar instances are found in `name`, `lineage`, `rank`, `parent`. Another functions, `is_common_ancestor` and `get_lineage_dataframe`, does not have any type checking. Briefly, what is currently being done is that we try to cast the `taxid` argument as an `int`, and if there is a ValueError, then a log message is displayed and we return `None` or something. The problem here is that there will be no error if you cast a float as an int. What is more, the rounding in this case is also weird, so for example `12.9` would become `12` when cast as an int. I would suggest that since this check is done a few different times that a new function is made (perhaps called `is_valid_taxid`), that first checks that the input is an int and secondly that it is included in nodes.dmp or merged.dmp. Probably the best way of doing the type check is as follows:

```python
# Note you will have to insert code to check whether the taxid is in nodes.dmp or merged.dmp
if isinstance(taxid, int):
    return True
elif isinstance(taxid, float):
[...]

PR #66, opened 2020-05-05
fixes-#54 Metagenome

WiscEvan:
* resolves issue-#54
* :racehorse: Add cache for properties reading assembly sequences. (start to #53)
* :bug: `NCBI_DIR` constant update (start to #33)
* :memo: Update docstrings missing for functions.
* :racehorse: Add cache for properties reading assembly sequences.
* :art: Add Entrypoint functionality for incorporation to packaging/distribution.
* :art::bug: Update main() to handle updated functions.


autometa/common/mag.py
```diff
@@ -63,28 +65,53 @@ def __init__(self, assembly, contigs, outdir=None):
...
 
     @property
+    @lru_cache(maxsize=None)
+    def assembly_seqs(self):
+        with open(self.assembly) as fh:
+            return [seq for title,seq in SimpleFastaParser(fh)]
+
+    @property
+    @lru_cache(maxsize=None)
+    def mag_seqs(self):
```
jason-c-kwan, line 75: Is this function misnamed? I'm just wondering if you meant to call this `mg_seqs` because I'm not sure `mag_seqs` makes sense as a name for this.

autometa/common/metagenome.py
```diff
@@ -537,88 +595,140 @@ def write_ranks(self, rank='superkingdom'):
...
-    parser.add_argument('--kmer-size', default=5, type=int)
-    parser.add_argument(
-        '--kmer-normalized',
-        help='Perform CLR transform on k-mer frequencies if provided. (</path/to/kmers.normalized.tsv>)',
-    )
-    parser.add_argument(
-        '--kmer-fpath',
-        help='</path/to/kmers.tsv>',
-    )
+    This handles filtering by length and taxon as well as ORF calling,
```
jason-c-kwan, line 620: Should be "this **script**".

autometa/common/metagenome.py
```diff
@@ -537,88 +595,140 @@ def write_ranks(self, rank='superkingdom'):
...
+    and various metagenome statistics.
+    """,
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
jason-c-kwan, line 623: Do we want to use this formatter in the `template.py` script?

autometa/common/metagenome.py
```diff
@@ -537,88 +595,140 @@ def write_ranks(self, rank='superkingdom'):
...
+        '(Will write to path if it does not exist).',
+        default=None)
     # Eventually will need to create a subparser for the taxon assignment methods
     # to include help information and required parameters according to method.
-    parser.add_argument('--cpus', help='num cpus to use', default=0)
-    parser.add_argument('--noparallel', help="Do not use GNU parallel",
-        action='store_false', default=True)
-    parser.add_argument('--force', help="overwrite existing files",
+    parser.add_argument('--cpus', help='Number of cpus to use when performing '
+        'annotations (Default will use all possible if `--parallel` is supplied).',
```
jason-c-kwan, line 665: Is this correct syntax? I mean the quote at the end of the line, and another quote directly after the newline. Seems like something that will cause a syntax error when run by Python.

WiscEvan: Since this is within the grouping `(` python will parse this as one string if a new line is encountered.

You can test this behavior with this example:


```python
# Success
line = ('here is a sentence containing multiple '
        'lines to display as one line implicitly')
print(line)

# Fails
line = 'here is a sentence containing multiple '
        'lines to display as one line implicitly'
print(line)

```

autometa/common/metagenome.py
WiscEvan: This is only required in scripts where we have defaults for parameters specified within parser parameters. Otherwise this is unnecessary. Kind of a judgement call here. We could add it in for easy reference?

jason-c-kwan: Is there a consequence to having this formatter when we don't specify defaults? Because I feel if there is no consequence, then adding it might help remove one gotcha.

autometa/common/metagenome.py
jason-c-kwan: Wow, I never knew this. Is this specific to Python 3?

WiscEvan: It appears to work with python2.7 as well. I'm not sure when this became possible

autometa/common/metagenome.py
WiscEvan: I'll add it to the template. As far as I know, there is not really any consequence to having it as the default formatter.

WiscEvan: The only consequence I can tell is the help text becomes a bit more verbose with optional parameters that do not have a default resulting in `(default: None)`.

e.g.

```shell
python docs/template.py -h
usage: template.py [-h] [--optional OPTIONAL] [--hello-world HELLO_WORLD]
                   positional

Concise description of script.

positional arguments:
  positional            <help text of positional arg>

optional arguments:
  -h, --help            show this help message and exit
  --optional OPTIONAL   <help text of optional arg> (default: None)
  --hello-world HELLO_WORLD
                        <help text of hello world> (default: Hello World)
```

jason-c-kwan: Well, I will leave it up to you. Perhaps it would be nice to include a comment in the template about the options?

WiscEvan: That's a good point. I'll put it in and comment on the subtle difference.

PR #67, opened 2020-05-05
Update MAG class to MetaBin

### LINKED ISSUE #53: [dev branch] Issues with autometa/common/mag.py

WiscEvan:
* resolves #53
* :art: Update to pass coverage filepath in `user.py` for `get_binning` rather than `pd.DataFrame`.
* :memo: Add docstrings to methods and class.
* :art::fire: Remove split_nucleotides method.
* :art: Update get_binning function to handle coverages as filepath not pd.DataFrame.
* :art: Change `mag.py` to `metabin.py`.
* :racehorse: fixed from PR-#66. Add cache functionality to time-consuming property methods.
* :memo: Add COMBAK comment for checkpointing. Return to this when implemented in `utilities.py`.


autometa/common/metabin.py
```diff
@@ -127,70 +285,73 @@ def get_orfs(self, orf_type='prot'):
...
-Mean GC content: {self.gc_content:4.2f}%
-________________________
-''')
+        print('Metabin Details\n'
+            '________________________\n'
+            f'Num. Contigs: {self.nseqs:,} / {self.nallseqs:,} ({self.seqs_pct:4.2f}% of total metagenome)\n'
+            f'Num. Nucl. ORFs: {self.nnucls:,}\n'
+            f'Num. Prot. ORFs: {self.nprots:,}\n'
+            f'Size: {self.size:,}bp / {self.totalsize:,}bp ({self.size_pct:4.2f}% of total metagenome)\n'
+            f'Mean GC content: {self.length_weighted_gc:4.2f}%\n'
```
jason-c-kwan, line 335: The print output should probably explicitly say the GC content is length-weighted.

WiscEvan:
I recommend addressing #66 prior to this.

jason-c-kwan:
The comment above about the `describe` output of the `MetaBin` class still needs to be addressed.

WiscEvan:
This was addressed in commit https://github.com/KwanLab/Autometa/pull/67/commits/3033b0d1f526ddb4c07919fccc7545fdc711f905
 
The changes should be reflected in the `files changed` section. You may need to refresh your page. Otherwise, the code is linked below:

https://github.com/WiscEvan/Autometa/blob/3033b0d1f526ddb4c07919fccc7545fdc711f905/autometa/common/metabin.py#L319-L345

PR #70, opened 2020-05-06
:art::bug: Add mock import of modules and link to contribution guidelines (fixes #22)

### LINKED ISSUE #22: [Dev branch] Document update needed
For the 2.0 release, we need to get the documentation up to date and significantly more friendly for non-technical people to understand. The quality of documentation goes a long way to increasing how much an application is used, which also increases citations for us and general interest. 

I discussed this briefly with @WiscEvan and it would probably be best to use [https://readthedocs.org/](readthedocs), because the information we want to include is probably too much for one README file. This has the advantage that it has versioning and is tied to the github repo. So you would update the docs in the same pull request as your feature update.

Some things that the docs should include:

* Detailed installation instructions, including using Docker and conda (when available)

* Step by step instructions of what to do with an assembly to get bins, and the rationale for choosing different options/parameters.

* How to critically assess the results with R or the web interface (when available). 

Note: I think this is a good issue for someone to work on if they are not yet confident enough in coding to tackle the other issues, and writing documentation is a good way to get an understanding of the pipeline. The language should be friendly and straight forward as possible.

Sidduppal:
- :bug: Added mock imports in `conf.py`, eliminating the need of `requirnments.txt`
- :art: Added [link](https://github.com/KwanLab/Autometa/blob/master/.github/CONTRIBUTING.md]) to contribution guidelines
- Resolves #22


autometa/common/markers.py
```diff
@@ -255,18 +255,18 @@ def main():
         datefmt='%m/%d/%Y %I:%M:%S %p',
         level=logger.DEBUG)
     parser = argparse.ArgumentParser(
-        description='Annotate ORFs with kingdom-marker information',
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
WiscEvan, line 259: The `formatter_class` should not be deleted.

autometa/common/markers.py
```diff
@@ -255,18 +255,18 @@ def main():
...
+        description='Annotate ORFs with kingdom-marker information')
```
WiscEvan, line 258:
```suggestion
        description='Annotate ORFs with kingdom-marker information',
        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```

PR #72, opened 2020-05-07
hmmer

### LINKED ISSUE #43: [dev branch] Issues with autometa/common/external/hmmer.py
This file contains functions required for running hmmer on metagenome sequences. I noticed the following issues with the script:

1. The script does not conform to the current `template.py` script, because the `main` function takes arguments. The code currently in the `if __name__ == '__main__':` block should be moved to `main`.

2. In the `main` function, we run the `hmmscan` function in a try/except block. However because the `force` command line argument is passed to the function, I'm not sure if FileExistsError will ever be encountered here. We should perhaps handle any other exception here.

3. In the `hmmscan` function there is the option of running hmmscan with either GNU parallel, which splits up the input and runs separate processes, or else to run hmmscan normally but specifying a number of CPUs to use. When using GNU parallel, a temp directory is made and at the end of the process, the separate output files for each process are combined. The problem is that as currently written, the contents of all files is copied to memory before being written to disk:

https://github.com/KwanLab/Autometa/blob/04f25f4b0bcf95fd6daa293ebd3f2461671c3f30/autometa/common/external/hmmer.py#L115-L122

It would be more memory efficient to open the output filehandle, then write lines to it directly as they are being read from each output file. That way, no more than one line has to be in memory at one time.

```python
out = open(outfpath, 'w')
for fp in tmpfpaths:
    with open(fp) as fh:
        for line in fh:
            out.write(line)
out.close()
```

Sidduppal:
- :art::racehorse: Use tempfile module
- :art: write lines as they are read
- Resolves #43
- :art: code currently in the `if __name__ == __main__:` block moved to main
- :racehorse: Output filehandle writes lines directly as they are being read from temporary file
- :art::racehorse: Replaced manual creation of temporary files with tempfile module
- Did not remove `FileExistsError` from the `except` block of `main` as that will be needed if `--force` is not passed. I could not identify any other error to catch/raise. 

:bug: Suggestion:
Current code for checking error while running hmmscan function:
```python
    if parallel:
        returncode = subprocess.call(' '.join(cmd), shell=True)
        tmp_fpaths = glob(os.path.join(tmp_dirpath, '*.txt'))
        with open(outfpath, 'w') as out:
            for fp in tmp_fpaths:
                with open(fp) as fh:
                    for line in fh:
                        out.write(line)
        shutil.rmtree(tmp_dirpath)
    else:
        with open(os.devnull, 'w') as STDOUT, open(os.devnull, 'w') as STDERR:
            proc = subprocess.run(cmd, stdout=STDOUT, stderr=STDERR)
        returncode = proc.returncode
    if returncode == 141:
        logger.warning(f'Make sure your hmm profiles are pressed!')
        logger.warning(f'hmmpress -f {hmmdb}')
        logger.error(f'Args:{cmd} ReturnCode:{returncode}')
        raise OSError(f'hmmscan: Args:{cmd} ReturnCode:{returncode}')
    if not os.path.exists(outfpath):
        raise OSError(f'{outfpath} not written.')
    return outfpath
```
I was using hmm database with was not hmmpressed and did not got an error. Suggested change:
```python
    if parallel:
        proc = subprocess.run(' '.join(cmd), shell=True)
        tmp_fpaths = glob(os.path.join(tmp_dirpath, '*.txt'))
        with open(outfpath, 'w') as out:
            for fp in tmp_fpaths:
                with open(fp) as fh:
                    for line in fh:
                        out.write(line)
        shutil.rmtree(tmp_dirpath)
    else:
        with open(os.devnull, 'w') as STDOUT, open(os.devnull, 'w') as STDERR:
            proc = subprocess.run(cmd, stdout=STDOUT, stderr=STDERR)
    try:
        proc.check_returncode()
    except subprocess.CalledProcessError as err:
        logger.warning(f'Make sure your hmm profiles are pressed!')
        logger.warning(f'hmmpress -f {hmmdb}')
        logger.error(f'Args:{cmd} ReturnCode:{proc.returncode}')
        raise err
    if not os.path.exists(outfpath):
        raise OSError(f'{outfpath} not written.')
    return outfpath
```

### DIFF
--- a/autometa/common/external/hmmer.py
+++ b/autometa/common/external/hmmer.py
@@ -26,8 +26,9 @@ Functions related to running hmmer on metagenome sequences
 
 import logging
 import os
-import subprocess
 import shutil
+import subprocess
+import sys
 import tempfile
 
 import pandas as pd
@@ -110,23 +111,30 @@ def hmmscan(orfs, hmmdb, outfpath, cpus=0, force=False, parallel=True, log=None)
                str(cpus), '--tblout', outfpath, hmmdb, orfs]
     logger.debug(f'cmd: {" ".join(cmd)}')
     if parallel:
-        returncode = subprocess.call(' '.join(cmd), shell=True)
+        proc = subprocess.run(' '.join(cmd), shell=True)
         tmp_fpaths = glob(os.path.join(tmp_dirpath, '*.txt'))
+        lines = ''
+        buffer_limit = 60000
         with open(outfpath, 'w') as out:
             for fp in tmp_fpaths:
                 with open(fp) as fh:
                     for line in fh:
-                        out.write(line)
+                        lines += line
+                        if sys.getsizeof(lines) >= buffer_limit:
+                            out.write(lines)
+                            lines = ''
+            out.write(lines)
         shutil.rmtree(tmp_dirpath)
     else:
-        with open(os.devnull, 'w') as STDOUT, open(os.devnull, 'w') as STDERR:
-            proc = subprocess.run(cmd, stdout=STDOUT, stderr=STDERR)
-        returncode = proc.returncode
-    if returncode == 141:
+        proc = subprocess.run(
+            cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    try:
+        proc.check_returncode()
+    except subprocess.CalledProcessError as err:
         logger.warning(f'Make sure your hmm profiles are pressed!')
         logger.warning(f'hmmpress -f {hmmdb}')
-        logger.error(f'Args:{cmd} ReturnCode:{returncode}')
-        raise OSError(f'hmmscan: Args:{cmd} ReturnCode:{returncode}')
+        logger.error(f'Args:{cmd} ReturnCode:{proc.returncode}')
+        raise err
     if not os.path.exists(outfpath):
         raise OSError(f'{outfpath} not written.')
     return outfpath
@@ -254,18 +262,14 @@ def main():
 
     if args.verbose:
         logger.setLevel(logger.DEBUG)
-    try:
-        result = hmmscan(
-            orfs=args.orfs,
-            hmmdb=args.hmmdb,
-            outfpath=args.hmmscan,
-            cpus=args.cpus,
-            force=args.force,
-            parallel=args.parallel,
-            log=args.log)
-    except FileExistsError as err:
-        logger.debug(err)
-        result = args.hmmscan
+    result = hmmscan(
+        orfs=args.orfs,
+        hmmdb=args.hmmdb,
+        outfpath=args.hmmscan,
+        cpus=args.cpus,
+        force=args.force,
+        parallel=args.parallel,
+        log=args.log)
 
     result = filter_markers(
         infpath=result,
--- a/autometa/common/external/hmmer.py
+++ b/autometa/common/external/hmmer.py
@@ -136,7 +136,7 @@ def hmmscan(orfs, hmmdb, outfpath, cpus=0, force=False, parallel=True, log=None)
         logger.error(f'Args:{cmd} ReturnCode:{proc.returncode}')
         raise err
     if not os.path.exists(outfpath):
-        raise OSError(f'{outfpath} not written.')
+        raise FileNotFoundError(f'{outfpath} not written.')
     return outfpath
 
 
@@ -256,20 +256,23 @@ def main():
                         action='store_true')
     parser.add_argument('--cpus', help='num cpus to use', default=0, type=int)
     parser.add_argument(
-        '--parallel', help="Enable GNU parallel", action='store_true')
+        '--parallel', help="enable GNU parallel", action='store_true')
     parser.add_argument('--verbose', help="add verbosity", action='store_true')
     args = parser.parse_args()
 
     if args.verbose:
         logger.setLevel(logger.DEBUG)
-    result = hmmscan(
-        orfs=args.orfs,
-        hmmdb=args.hmmdb,
-        outfpath=args.hmmscan,
-        cpus=args.cpus,
-        force=args.force,
-        parallel=args.parallel,
-        log=args.log)
+    if os.path.exists(args.hmmscan) and os.stat(args.hmmscan).st_size > 0 and not args.force:
+        result = args.hmmscan
+    else:
+        result = hmmscan(
+            orfs=args.orfs,
+            hmmdb=args.hmmdb,
+            outfpath=args.hmmscan,
+            cpus=args.cpus,
+            force=args.force,
+            parallel=args.parallel,
+            log=args.log)
 
     result = filter_markers(
         infpath=result,
--- a/autometa/common/external/hmmer.py
+++ b/autometa/common/external/hmmer.py
@@ -76,7 +76,7 @@ def hmmscan(orfs, hmmdb, outfpath, cpus=0, force=False, parallel=True, log=None)
         hmmscan failed
     """
     # OPTIMIZE: we want to extend parallel to grid computing (workqueue?) via --sshlogin?
-    if os.path.exists(outfpath) and os.stat(outfpath).st_size > 0 and not force:
+    if os.path.exists(outfpath) and os.path.getsize(outfpath) > 0 and not force:
         raise FileExistsError(f'{outfpath}. Use force to overwrite!')
     if parallel:
         outdir = os.path.dirname(os.path.realpath(outfpath))
@@ -206,7 +206,7 @@ def filter_markers(infpath, outfpath, cutoffs, orfs=None, force=False):
     for fp in [infpath, cutoffs]:
         if not os.path.exists(fp):
             raise FileNotFoundError(f'{fp} not found')
-    if os.path.exists(outfpath) and os.stat(outfpath).st_size > 0 and not force:
+    if os.path.exists(outfpath) and os.path.getsize(outfpath) > 0 and not force:
         raise FileExistsError(f'{outfpath} already exists')
     hmmtab_header = ['sname', 'sacc', 'orf', 'score']
     col_indices = [0, 1, 2, 5]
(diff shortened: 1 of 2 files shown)


Sidduppal:
One caveat in my suggestion is that it is not specific to unpressed databases it will raise an error anytime `subprocess.run()` does not run porperly

autometa/common/external/hmmer.py
```diff
@@ -234,10 +230,30 @@ def translater(x): return translations.get(x, x.rsplit('_', 1)[0])
...
+        '--parallel', help="Enable GNU parallel", action='store_true')
+    parser.add_argument('--verbose', help="add verbosity", action='store_true')
+    args = parser.parse_args()
+
     if args.verbose:
         logger.setLevel(logger.DEBUG)
-
     try:
         result = hmmscan(
             orfs=args.orfs,
```
WiscEvan, line 259: >2. In the main function, we run the hmmscan function in a try/except block. However because the force command line argument is passed to the function, I'm not sure if FileExistsError will ever be encountered here. We should perhaps handle any other exception here.

Has this been addressed?

WiscEvan:
In the future, it would also be better if you submitted the PR from a separate branch than `dev`. Right now I do not think you have an easy method of delineating your `dev` branch from the `KwanLab` until this PR is addressed.

autometa/common/external/hmmer.py
Sidduppal: I actually commented about in the PR above. Copying from above,
_"Did not remove FileExistsError from the except block of main as that will be needed if --force is not passed. I could not identify any other error to catch/raise."_

Sidduppal:
2 questions before I make the commit
- Strangely, when I ran the hmmscan function, it was way way faster in `parallel` than when I did `cpus=30`. I did this using the original `hmmer.py` script. Do you still want me to remove the GNU parallel option?
- In this [comment](https://github.com/KwanLab/Autometa/issues/43#issuecomment-626203264) can you suggest an appropriate value for the `buffer_limit` or point me towards some helpful resources? I tried to look about it but could not find anything useful

autometa/common/external/hmmer.py
jason-c-kwan: What I meant was that the `hmmscan` function has error handling for `FileExistsError`, so this error will never be encountered by `main`.

autometa/common/external/hmmer.py
```diff
@@ -79,11 +80,9 @@ def hmmscan(orfs, hmmdb, outfpath, cpus=0, force=False, parallel=True, log=None)
     if parallel:
```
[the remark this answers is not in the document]
jason-c-kwan: What was the difference in time, and how many CPUs did you use? Below @Sidduppal seems to have had different results. If the difference is negligible, I would favor simplifying with just one option.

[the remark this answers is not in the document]
WiscEvan: ### 1st Run w/o GNU parallel and 4 cpus

```shell
time python -m autometa.common.external.hmmer --cpus 4 --force tests/data/metagenome.orfs.faa autometa/databases/markers/bacteria.single_copy.hmm autometa/databases/markers/bacteria.single_copy.cutoffs tests/data/metagenome.hmmscan.out tests/data/metagenome.markers.tsv

142.50s user 73.71s system 105% cpu 3:25.45 total
```

```shell
time python -m autometa.common.external.hmmer --cpus 4 --force tests/data/metagenome.orfs.faa autometa/databases/markers/bacteria.single_copy.hmm autometa/databases/markers/bacteria.single_copy.cutoffs tests/data/metagenome.hmmscan.out tests/data/metagenome.markers.tsv

143.17s user 75.06s system 104% cpu 3:28.19 total
```

### 2nd Run with GNU parallel and 4 cpus
```shell
time python -m autometa.common.external.hmmer --cpus 4 --parallel --force tests/data/metagenome.orfs.faa autometa/databases/markers/bacteria.single_copy.hmm autometa/databases/markers/bacteria.single_copy.cutoffs tests/data/metagenome.hmmscan.out tests/data/metagenome.markers.tsv

166.16s user 105.19s system 378% cpu 1:11.72 total
```

### 3rd Run with GNU parallel and all cpus available
```shell
time python -m autometa.common.external.hmmer --parallel --force tests/data/metagenome.orfs.faa autometa/databases/markers/bacteria.single_copy.hmm autometa/databases/markers/bacteria.single_copy.cutoffs tests/data/metagenome.hmmscan.out tests/data/metagenome.markers.tsv

230.71s user 148.27s system 703% cpu 53.875 total
```

### 4th Run w/o GNU parallel and explicitly 8 cpus

```shell
time python -m autometa.common.external.hmmer --cpus 8 --force tests/data/metagenome.orfs.faa autometa/databases/markers/bacteria.single_copy.hmm autometa/databases/markers/bacteria.single_copy.cutoffs tests/data/metagenome.hmmscan.out tests/data/metagenome.markers.tsv

156.35s user 110.63s system 118% cpu 3:45.22 total
```

### 5th Run w/o GNU parallel and "0" cpus
```shell
time python -m autometa.common.external.hmmer --force tests/data/metagenome.orfs.faa autometa/databases/markers/bacteria.single_copy.hmm autometa/databases/markers/bacteria.single_copy.cutoffs tests/data/metagenome.hmmscann.out tests/data/metagenome.markers.tsv

123.77s user 63.02s system 99% cpu 3:07.67 total
```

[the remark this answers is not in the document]
jason-c-kwan: I don't know what to make of these results, but my intuition is that it will probably make more difference for very large files. Perhaps we should just use whatever is simplest to implement for now?

autometa/common/external/hmmer.py
jason-c-kwan: So, no @Sidduppal it is still not addressed.

Sidduppal: I am not sure about that, because `hmmscan` is just rasing the error not handling it.

autometa/common/external/hmmer.py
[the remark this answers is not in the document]
WiscEvan: I think we will notice a bigger difference with larger files as well. I think it is safest to leave both as an option.

autometa/common/external/hmmer.py
jason-c-kwan: The `hmmscan` function raises a `FileExistsError` when the output exists and the user has not specified `force = True`. That means the program will exit. I don't know what you mean by handling it? The user has the choice to specify `force` or not. If you just go ahead and overwrite that is choosing for them (you may as well not have that argument), and the only alternative is stopping the program and informing the user why. My complaint is that as written, since this error is thrown by `hmmer`, then `main` will never encounter a `FileExistsError`, so that try/except block should be taken out. Leaving it in is unnecessarily confusing to developers. Say for instance someone in the future wants to alter the error handling here when running through command line. If they change `main` they will be confused why the behavior doesn't change. The answer is they were looking in the wrong place!

autometa/common/external/hmmer.py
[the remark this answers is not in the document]
jason-c-kwan: OK, let's leave both in for now.

autometa/common/external/hmmer.py
```diff
@@ -111,26 +111,30 @@ def hmmscan(orfs, hmmdb, outfpath, cpus=0, force=False, parallel=True, log=None)
...
+        proc.check_returncode()
+    except subprocess.CalledProcessError as err:
         logger.warning(f'Make sure your hmm profiles are pressed!')
         logger.warning(f'hmmpress -f {hmmdb}')
-        logger.error(f'Args:{cmd} ReturnCode:{returncode}')
-        raise OSError(f'hmmscan: Args:{cmd} ReturnCode:{returncode}')
+        logger.error(f'Args:{cmd} ReturnCode:{proc.returncode}')
+        raise err
     if not os.path.exists(outfpath):
         raise OSError(f'{outfpath} not written.')
```
WiscEvan, line 139: This should probably be a `FileNotFoundError`, since it is checking whether `outfpath` exists.

```suggestion
        raise FileNotFoundError(f'{outfpath} not written.')
```

autometa/common/external/hmmer.py
```diff
@@ -280,4 +259,25 @@ def main(args):
...
+        force=args.force,
+        parallel=args.parallel,
+        log=args.log)
+
+    result = filter_markers(
+        infpath=result,
+        outfpath=args.markers,
+        cutoffs=args.cutoffs,
+        orfs=args.orfs,
+        force=args.force)
```
WiscEvan, line 279: How do we account for when a user is supplying hmmscan results and just want to filter the results given the cutoffs? 

### Expected behavior:

Filter results given a previously generated hmmscan table

### Scenario

1. I have hmmscan results from the bacteria single-copy marker genes.
2. I do **not** want to overwrite these results.
3. I want to filter them based on the respective cutoffs.

### **Error encountered:** 
When I run `hmmer.py` I receive `FileExistsError` for my hmmscan results and the program exits. 
Can I filter these results without having to overwrite my previously generated hmmscan table?

autometa/common/external/hmmer.py
```diff
@@ -234,11 +238,33 @@ def translater(x): return translations.get(x, x.rsplit('_', 1)[0])
...
+    parser.add_argument(
+        '--parallel', help="enable GNU parallel", action='store_true')
+    parser.add_argument('--verbose', help="add verbosity", action='store_true')
+    args = parser.parse_args()
+
     if args.verbose:
         logger.setLevel(logger.DEBUG)
-
-    try:
+    if os.path.exists(args.hmmscan) and os.stat(args.hmmscan).st_size > 0 and not args.force:
```
WiscEvan, line 265: This looks like it will resolve the issue. I've noticed we can also get the file size with `os.path.getsize(fpath)` which I think is a bit more intuitive than `os.stat(fpath).st_size`. Minor nitpick, but I recently came across this and have been changing this as I come across it. I think this occurs a few times in this script. So feel free to update throughout.

```suggestion
    if os.path.exists(args.hmmscan) and os.path.getsize(args.hmmscan) > 0 and not args.force:
```

PR #76, opened 2020-05-09
:art: :racehorse: Resolves #55

### LINKED ISSUE #55: [dev branch] Issues with autometa/config/environ.py
This file is a collection of functions useful for the configuration of the Autometa environment. I found the following issues:

1. In the `which` function, the docstring should mention that the function checks that the `program` argument (a path) is executable, not just has read permissions. Note also that the docstring says that the function returns `None` if the path is not valid, when in actual fact it returns an empty string. These two things are not the same, and you can see in the Python prompt that `None == ''` returns `False`. So the return value should match the docstring so that client code expects the right thing.

2. In the `get_versions` function, we set up a `dispatcher` dictionary:

https://github.com/KwanLab/Autometa/blob/3e8c7718dc63e5a06654a899506535ae8bd6219e/autometa/config/environ.py#L262-L272

What this is doing is it is storing functions used to find program versions, under the relevant key which is the string value of that program. So, obviously this works, but at the top of the file we already have a constant called `EXECUTABLES`, which stores all the names of the dependency executables:

https://github.com/KwanLab/Autometa/blob/3e8c7718dc63e5a06654a899506535ae8bd6219e/autometa/config/environ.py#L42-L51

My problem here is that the `dispatcher` variable is unnecessary, and it would be much better to have the list of dependencies just defined in one place to avoid errors when it is updated. This page: ttps://stackoverflow.com/questions/3061/calling-a-function-of-a-module-by-using-its-name-a-string outlines ways to call functions by the string of their name. I would suggest something like this:

```python
for exe, found in executables.items():
    if found:
        version = locals()[exe]()
    else:
        ...
```

3. Another problem with the `get_version` function is the check whether the argument `program` is a string:

https://github.com/KwanLab/Autometa/blob/3e8c7718dc63e5a06654a899506535ae8bd6219e/autometa/config/environ.py#L273-L276

Notice that the `if type(program) is not str:` statement is underneath `exe_name = os.path.basename(program)`? Well, if we try to run `os.path.basename` on a number or other non-str type then we will get a TypeError, so the type check should be before this statement.

Sidduppal:
- :art: :racehorse: Replaced dispatcher dict by `globals()` [See](https://docs.python.org/3/library/functions.html#globals)
- :art: Changed the location of the checking `TypeError` in `getversions()` function
- :art: :fire: Suggestion: Current code to check if a program is in a the `$PATH` and is executable, is
```python
def which(program):
    def is_exe(fpath):
        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)
    # check to see if program path is given or just the name
    fpath, __ = os.path.split(program)
    if fpath:
        if is_exe(program):
            return program
    else:
        for path in os.environ["PATH"].split(os.pathsep):
            exe_file = os.path.join(path, program)
            if is_exe(exe_file):
                return exe_file
    return ''
``` 
We can use `shutil.which()` instead, which is concise and removes unneccesary code. [See](https://docs.python.org/3/library/shutil.html#shutil.which). This does exactly what the above code does and can take program name or program path as input.
Recommendation:
```python
def which(program, mode=os.X_OK):
     return shutil.which(program)
```
The above code will return the file path if the program is in `$PATH` and is executable and None if not. The return is slightly different from what the original code returns, where if the program is not in the `$PATH` or is not an executable it returns an empty string `''`. From my reading of the code returning `None` should not cause any errors, however if we do want to return an empty string in case the program is not in `$PATH` or not executable then we can use this:
```python
def which(program, mode=os.X_OK):
    fpath = shutil.which(program)
    if fpath:
        return fpath
    return ''
```

### DIFF
--- a/autometa/config/environ.py
+++ b/autometa/config/environ.py
@@ -52,25 +52,6 @@ EXECUTABLES = [
 ]
 
 
-def which(program, mode=os.X_OK):
-    """
-    Finds the full path for an executable and checks read permissions exist
-
-    Parameters
-    ----------
-    program : str
-        the program to check
-    mode : path like object, optional
-        permission mask passed to os.access(), by default os.X_OK
-
-    Returns
-    -------
-    str or None
-        the path if it was valid, eg. </path/to/executable/> or None if not
-    """
-    return shutil.which(program)
-
-
 def find_executables():
     """Retrieves executable file paths by looking in Autometa dependent executables.
 
@@ -80,7 +61,7 @@ def find_executables():
         {executable:</path/to/executable>, ...}
 
     """
-    return {exe: which(exe) for exe in EXECUTABLES}
+    return {exe: shutil.which(exe, mode=os.X_OK) for exe in EXECUTABLES}
 
 
 def diamond():
@@ -92,7 +73,7 @@ def diamond():
         version of diamond
 
     """
-    exe = which('diamond')
+    exe = shutil.which('diamond', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe, 'version'],
         stdout=subprocess.PIPE,
@@ -110,7 +91,7 @@ def hmmsearch():
     str
         version of hmmsearch
     """
-    exe = which('hmmsearch')
+    exe = shutil.which('hmmsearch', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe, '-h'],
         stdout=subprocess.PIPE,
@@ -129,7 +110,7 @@ def hmmpress():
     str
         version of hmmpress
     """
-    exe = which('hmmpress')
+    exe = shutil.which('hmmpress', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe, '-h'],
         stdout=subprocess.PIPE,
@@ -148,7 +129,7 @@ def hmmscan():
     str
         version of hmmscan
     """
-    exe = which('hmmscan')
+    exe = shutil.which('hmmscan', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe, '-h'],
         stdout=subprocess.PIPE,
@@ -167,7 +148,7 @@ def prodigal():
     str
         version of prodigal
     """
-    exe = which('prodigal')
+    exe = shutil.which('prodigal', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe, '-v'],
         stdout=subprocess.DEVNULL,
@@ -185,7 +166,7 @@ def bowtie2():
     str
         version of bowtie2
     """
-    exe = which('bowtie2')
+    exe = shutil.which('bowtie2', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe, '--version'],
         stdout=subprocess.PIPE,
@@ -203,7 +184,7 @@ def samtools():
     str
         version of samtools
     """
-    exe = which('samtools')
+    exe = shutil.which('samtools', mode=os.X_OK)
     proc = subprocess.Popen(
         [exe], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     stdout, stderr = proc.communicate()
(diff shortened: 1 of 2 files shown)


Sidduppal:
PS. Apologies for so many commits, I was not aware that new features should be worked on a different branch and since I worked on #55 and #43 on the same branch I was not able to create a new PR and all my future commits were adding into the PR related to #43 . Learned my lesson in the amount of effort and time wasted, will take care of it in the future :)

WiscEvan:
I think using `shutil.which` instead of the defined function `which` is a great idea. Reduces the amount of code and makes reading just a bit easier. I think the entire function definition can be removed, and you can replace `which` with `shutil.which` wherever it is called.

autometa/config/environ.py
```diff
@@ -269,36 +272,26 @@ def get_versions(program=None):
...
-        exe_name = os.path.basename(program)
         if type(program) is not str:
-            raise ValueError(f'program is not string. given:{type(program)}')
-        if exe_name not in dispatcher:
+            raise TypeError(f'program is not string. given:{type(program)}')
+        exe_name = os.path.basename(program)
+        if exe_name not in globals():
             raise KeyError(f'{exe_name} not in executables')
-        return dispatcher[exe_name]()
+        return (globals()[exe_name]())
```
WiscEvan, line 282: Do we need to wrap the value being returned with more parentheses?

autometa/config/environ.py
```diff
@@ -269,36 +272,26 @@ def get_versions(program=None):
...
-            version = dispatcher[exe]()
+            # globals()[exe]() accesses the location of the function for that program
+            # eg. location of bedtools(), prodigal(), etc..
+            version = globals()[exe]()
         else:
             logger.warning(f'VersionUnavailable {exe}')
             version = 'ExecutableNotFound'
         versions.update({exe: version})
-    return versions
+    return (versions)
```
WiscEvan, line 294: Do we need to wrap the returned value with parentheses?

autometa/config/environ.py
```diff
@@ -269,36 +272,26 @@ def get_versions(program=None):
...
-        return dispatcher[exe_name]()
+        return (globals()[exe_name]())
     versions = {}
     executables = find_executables()
     for exe, found in executables.items():
         if found:
-            version = dispatcher[exe]()
+            # globals()[exe]() accesses the location of the function for that program
+            # eg. location of bedtools(), prodigal(), etc..
+            version = globals()[exe]()
```
WiscEvan, line 289: Maybe we should split this into two lines for readability? 

I'm thinking something like:

```suggestion
            get_version = globals()[exe]
            version = get_version()
```

Perhaps this would be easier to understand for the future reader?

Sidduppal:
`shutil.which` has been added in the last commit :smile:

jason-c-kwan:
I can't see the commit here though.

autometa/config/environ.py
```diff
@@ -51,42 +52,23 @@
...
-    if fpath:
-        if is_exe(program):
-            return program
-    else:
-        for path in os.environ["PATH"].split(os.pathsep):
-            exe_file = os.path.join(path, program)
-            if is_exe(exe_file):
-                return exe_file
-    return ''
+    return shutil.which(program)
```
Sidduppal, line 71: I am not sure, why this is not showing up for you. I can commit again if you like. 
@WiscEvan are you also not able to see the last commit?

jason-c-kwan: Perhaps I am not understanding what you are trying to do. I expected there to be no `which` function at all, just a call to `shutil.which(program)` wherever there would have been a call to `which` before. Is that not what you are trying to do?

Sidduppal: It seems like I had misunderstood what you and @WiscEvan were trying to convey. I have completely removed the `which` function now and have now called `shutil.which` wherever `which` function was being called.

Issue #59, opened 2020-05-01
[dev branch] Issues with autometa/config/databases.py


jason-c-kwan:
This file contains configuration handling for Autometa databases. Here are the issues I noticed:

1. [x] The main docstring should note that the file contains the definition of the `Databases` class.

2. [x] The `Databases` class does not really have a docstring.

3. [x] There is no docstring for the `satisfied` property of the `Databases` class.

4. [x] In the docstring of the `prepare_sections` method, it would be useful if it explicitly says that `self.config` is amended, like in other method docstrings in this file. **(This function has now been removed and the routine moved to the `__init__`)**

5. [x] The `format_nr` method calls `diamond.makedatabase` to convert `nr.gz` into `nr.dmnd`. Currently `diamond.makedatabase` does not create a checksum file for the dmnd file after that task is completed successfully. That logic could probably be inserted into `format_nr`, so that later we can tell that the diamond database creation finished successfully and we can tell if the file became corrupted since then.

6. [x] In the `extract_taxdump` method, the `untar` function from `utilities.py` is called to extract individual files from `taxdump.tar.gz`. This function doesn't create a checksum file, so it would be good to insert that in `extract_taxdump`. This is so that later it can be checked that the various `.dmp` files are still OK. (NCBI I think supplies a checksum file for taxdump, but not the individual .dmp files as far as I know).

7. [x] In the `update_ncbi` method, the docstring does not specify what type of object the method expects the argument `options` to be. This would be useful for the documentation. **(`update_ncbi` has been removed and functionality is placed in `download_ncbi_files`)**

8. [x] In the `update_ncbi` method, URLs for downloading files from NCBI are taken from `DEFAULT_CONFIG` (i.e. whatever is in the user supplied config file is ignored). I think this is bad because a user might reasonably expect that changes in their config file to be reflected in behavior. There are legitimate reasons to do this as well - for example perhaps they need to use a NCBI FTP mirror website if do not have a fast connection to the main NCBI FTP site or if it is down for whatever reason. **(`update_ncbi` has been removed and functionality is placed in `download_ncbi_files`)**

9. [x] In the `update_ncbi` method, files are only downloaded if it is found that they don't already exist. Apart from the fact that we haven't checked that existing files are still OK (see other comments), or that the files on the NCBI FTP site are newer, since 'update' is in the name of the method, a user might reasonably expect that existing files will be updated/overwritten. **(`update_ncbi` has been removed and functionality is placed in `download_ncbi_files`)**

[...]

Issue #40, opened 2020-04-09
[dev branch] Issues with autometa/common/utilities.py


jason-c-kwan:
This file contains common utility functions (no classes) used by Autometa scripts. I found the following issues.

1. [x] This script does not follow the template in `docs/template.py`. In particular, there is no way to run any of the functions from the command line.

2. [x] The `gunzip` function is supposed to uncompress a `.gz` file. The problem is that it loads the entire thing into memory, which is not very efficient, and would perhaps fail for very large files. You only need to load a single line at a time, like this:

```python
output_file = open('output_path', 'w')
with gzip.open('input_path', 'rb') as fh:
    for line in fh:
        output_file.write(line)
output_file.close()
```

Note that if you use `gzip.open` in `'rb'` mode, then you don't need `line.decode()`. 

Another thing to note with this function is that it will more than double the space being taken up by the original `'.gz'` file. Perhaps there should be the option to delete the original `'.gz'` file, if the uncompressing finished correctly.

3. [x] The `file_length` function should work correctly, but it will take a very long time for some of the NCBI database files. Essentially what this function is used for is to return the number of lines in a file. Now, that is all well and good, but oftentimes in the pipeline, this is only used to get a figure for use by [tqdm](https://tqdm.github.io/). Tqdm is a module that provides nice looking progress bars for lengthy iterations, but it needs to know ahead of time how many iterations there will be. Why am I bringing this up? Well there are times in the pipeline where we use this function to get that figure, and it will take a long time because it has to go through the entire file. That is all well and good when we really need an accurate value, but when we want it for tqdm it is just for the sake of looking pretty! (In other words, we get a performance hit). So, I would suggest that there be a way to run this in "approximation mode". Sometimes the NCBI files are like tables where the lines vary a bit in length but are basically the same all the way through. One could make an estimate of the average line length, then use the total size of the file to estimate line number (e.g. by working out how much binary space should be taken up by lines of your average length).

4. [x] In the `get_checksum` function, the docstring incorrectly states that a checksum is retrieved for the provided `args`, but the one argument is `fpath`. Another issue is the use of the SHA-256 hashing algorithm here. SHA-1 would be much quicker, especially for very large files. SHA-1 is not used for security any more because it has been cracked. I would argue that in this case we don't need security and it would be better to use a fast hashing algorithm. I believe that SHA-1 is the fastest.

WiscEvan:
[...]

Issue #71, opened 2020-05-07
[dev branch] Checksumming/checkpointing needed to assure integrity of output files.


jason-c-kwan:
During the code review, it became clear that we need a general method of determining not just that an output file exists but that it is not corrupted in some way. This has been acknowledged by @WiscEvan in a few PR requests so far, but I thought I'd flag it as an issue here. This is important especially for time consuming and/or computationally intensive steps because there are a lot of reasons why, for example, DIAMOND could crash half-way through writing its output file. I propose the following:

1. Whenever a file is written by Autometa, a checksum file is made. Because we don't necessarily have to worry that much about security, a fast hashing algorithm like SHA-1 could be used.
2. Whenever a file is required later, a simple check that the checksum file is there and non-zero would indicate that the step previously was completed.
3. Checking that the checksum in the checksum file matches to a re-calculated checksum of the output file will confirm that the file was not corrupted since writing. Secondarily, this confirms that the checksum file is not corrupted, and that the initial calculation of the checksum didn't fail in some way. 

Although some could argue that the above falls under the category of "nice to have, but not essential", I would argue that it would help diagnose or even reduce support requests.

WiscEvan:
So open for discussion, but I was thinking of checkpointing during a run in terms of _one_ `checkpoints.txt` file that consists of checksums with the respective step.

### Arguments for writing one file instead of multiple:

1. A user may move their file and not the respective md5, which would result in not finding the respective md5 and a performance hit due to having to falsely re-do the annotation.

2. Progress reporting: If multiple annotations are located in multiple directories, searching for the respective md5 files may become even more cumbersome.

>1. One checkpoints.txt file would eliminate a user having to shuttle both the md5 and their file around.

>2. Inspecting a `checkpoints.txt`/`progress.txt`/`steps.txt`/`pipeline.txt` file would be easier than `globbing` for multiple file paths that may be in different directories.

### `checkpoints.txt` file structure:

Structure for writing/lookup in `checkpoints.txt`
#### <checkpoints.txt>
```shell
# Step successful
# <checksum><whitespace><filename>\n
# or
# <checksum><whitespace><function_called>\n

# Step not yet encountered or failed.
# <whitespace><whitespace><filename>\n
```

#### <checkpoints.txt> example
```shell
3fda9608a77b3809141f75a6069b4abd  length_filter
852011c3716e19e39a5e30be948896f5 call_orfs
[...]

PR #77, opened 2020-05-13
databases and utilities

WiscEvan:
- :art::memo: fixes issue-#59
- :art::memo::racehorse: Resolves issue-#40
- :art: Update .gitignore
- :art::racehorse: Add md5 checksums for markers database files
- :art: Update default.config with md5 checksum urls
- :racehorse: Update file_length functionality with approximate parameter


autometa/common/utilities.py
```diff
@@ -298,14 +320,76 @@ def blockiter(fh, blocksize=65536):
...
+    FileNotFoundError
+        Provided `fpath` does not exist.
+
+    """
+    if not isinstance(fpath, str):
+        raise TypeError(type(fpath))
+    if not os.path.exists(fpath):
+        raise FileNotFoundError(fpath)
+    with open(fpath) as fh:
+        return fh.readline()
```
jason-c-kwan, line 360: This will return the checksum with a newline character at the end (if one is there). Is that what we want? Perhaps we should do `return fh.readline().rstrip()`?

autometa/common/utilities.py
```diff
@@ -122,17 +128,20 @@ def gunzip(infpath, outfpath):
...
+        for i,line in enumerate(fh):
+            lines += line
+            if i >= block_size:
+                out.write(lines)
+                lines = ''
     logger.debug(f'gunzipped {infpath} to {outfpath}')
+    if delete_original:
+        os.remove(infpath)
+        logger.debug(f'removed original file: {infpath}')
     return outfpath
```
jason-c-kwan, line 145: This function should call `write_checksum` so that the file it produces can be checked later.

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
             raise TypeError(f'config is not ConfigParser : {type(config)}')
-        if type(dryrun) is not bool:
-            raise TypeError(
-                f'dryrun must be True or False. type: {type(dryrun)}')
-
-        self.config = config
-        self.dryrun = dryrun
-        self.nproc = nproc
-        self.prepare_sections()
+        if not isinstance(dryrun, bool) and isins:
```
jason-c-kwan, line 104: I couldn't find `isins` in the Python documentation. Was this supposed to be an `isinstance` clause?

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+            ftp_fullpath = self.config.get('checksums', option)
+            chksum_fpath = ftp_fullpath.split(host)[-1]
+            with FTP(host) as ftp, tempfile.TemporaryFile() as fp:
+                ftp.login()
+                result = ftp.retrbinary(f'RETR {chksum_fpath}', fp.write)
+                if not result.startswith('226 Transfer complete'):
+                    raise ConnectionError(f'{chksum_fpath} download failed')
+                ftp.quit()
+                fp.seek(0)
+                checksum = fp.read().decode()
```
jason-c-kwan, line 187: Has this been tested? I'm just wondering if it results in the checksum string with a newline at the end, which might cause problems later. Also, if one URL doesn't have a newline, can we assume that all won't?

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+        elif section == 'markers':
+            url = self.config.get('checksums', option)
+            with requests.Session() as session:
+                resp = session.get(url)
+                if not resp.ok:
+                    raise ConnectionError(f'Failed to retrieve {url}')
+                checksum = resp.text
```
jason-c-kwan, line 194: If the user supplies their own marker databases will this cause an error? The code looks different to the above, I guess assuming that the URL is not an FTP site. What if the user supplies a local file or a FTP address and/or doesn't provide a checksum entry in their config?

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+        return checksum
 
     def format_nr(self):
-        """Format NCBI nr.gz database into diamond formatted database nr.dmnd.
+        """Construct a diamond formatted database (nr.dmnd) from `nr` option
+        in `ncbi` section in user config.
+
+        NOTE: The checksum 'nr.gz.md5' will only be generated if nr.dmnd
+        construction is successful. If the provided `nr` option in `ncbi` is 'nr.gz'
+        the database will be removed after successful database formatting.
```
jason-c-kwan, line 203: This is sort of confusing. Shouldn't we generate a md5 file of the `nr.dmnd` file?

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
-        db_outfpath = db_infpath.replace('.gz', '.dmnd')
-        if not self.dryrun and not os.path.exists(db_outfpath):
-            diamond.makedatabase(
-                fasta=nr, database=db_infpath, nproc=self.nproc)
+        db_infpath = self.config.get('ncbi','nr')
+        db_infpath_md5 = f'{db_infpath}.md5'
+        db_outfpath = db_infpath.replace('.gz','.dmnd')
+        checksums_match = False
+        if os.path.exists(db_infpath_md5) and os.path.exists(db_outfpath):
+            # If nr.dmnd.md5 exists, then we will check if nr.gz.md5 is matching
```
jason-c-kwan, line 216: See my comment above. The naming of these files is very confusing. Shouldn't it be `nr.dmnd.md5`?

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+            remote_hash, __ = remote_checksum.split()
+            if current_hash == remote_hash:
+                logger.debug(f'nr checksums match, skipping...')
+                checksums_match = True
+
+        no_update_db = os.path.exists(db_outfpath) and not self.do_update
+        if self.dryrun or no_update_db or checksums_match:
+            self.config.set('ncbi', 'nr', db_outfpath)
+            logger.debug(f'set ncbi nr: {db_outfpath}')
+            return
```
jason-c-kwan, line 230: After this point in the code we are here because `checksums_match == False`. So shouldn't we redownload the database? Or at least have some sort of option to do so. It seems weird that we check whether the checksums match and then don't do anything about it. Also, couldn't `checksums_match == True` but we still need to make the `nr.dmnd` file? It seems that in that scenario, the function will `return` and not format the database.

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
             outfpath = os.path.join(self.ncbi_dir, fname)
-            if not self.dryrun and not os.path.exists(outfpath):
-                outfpath = untar(taxdump, self.ncbi_dir, fname)
+            if self.dryrun:
+                logger.debug(f'UPDATE (ncbi,{option}): {outfpath}')
+                self.config.set('ncbi', option, outfpath)
+                continue
+            # Only update the taxdump files if the user says to do an update.
+            if self.do_update and os.path.exists(outfpath):
+                os.remove(outfpath)
```
jason-c-kwan, line 273: Here, wouldn't we need to also check that the `taxdump` file exists, because previous runs of this function would have deleted it? The below code assumes it exists.

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+            # Only extract the taxdump files if this is not a "dryrun"
+            if not os.path.exists(outfpath):
+                outfpath = untar(taxdump_fpath, self.ncbi_dir, fname)
             logger.debug(f'UPDATE (ncbi,{option}): {outfpath}')
             self.config.set('ncbi', option, outfpath)
+        if self.dryrun:
+            return
+        taxdump_md5 = f'{taxdump_fpath}.md5'
+        write_checksum(taxdump_fpath, taxdump_md5)
```
jason-c-kwan, line 282: This part is a little confusing to me. I get that making a md5 file of the `taxdump` file before deleting allows you to later check whether the version we have is the same as is found on the FTP server. But why is this not done when the file is downloaded (I mean calculated, to verify the download)? Wouldn't it make sense to also make checksums of the extracted files, so that later we can verify that they are not corrupted?

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+            if self.dryrun:
+                return
+
+            with FTP(host) as ftp, open(outfpath, 'wb') as fp:
+                ftp.login()
+                logger.debug(f'starting {option} download')
+                result = ftp.retrbinary(f'RETR {ftp_fpath}', fp.write)
+                if not result.startswith('226 Transfer complete'):
+                    raise ConnectionError(f'{option} download failed')
+                ftp.quit()
```
jason-c-kwan, line 327: Here after download, it would make sense to then calculate the md5 of the file and check it against the corresponding md5 URL on the FTP server. Writing the md5 to a file then could signify to the pipeline that the download was successful, and then checking the file later against the md5 file would assure that the file is not corrupted.

autometa/common/utilities.py
WiscEvan: To keep consistency with other checksum files, I return the full line with special characters. For example, retrieving an md5 from NCBI would also contain the newline character, which would be reflected in the checksum calculation, so this is included.

```python
In [1]: write_checksum("VERSION", 'version.md5')

In [2]: !cat version.md5
3fda9608a77b3809141f75a6069b4abd VERSION

In [3]: with open('VERSION') as fh:
    ...:     line = fh.readline()

In [4]: line
Out[4]: '2.0a0\n'

In [5]: line.strip()
Out[5]: '2.0a0'

In [6]: with open('test','w') as fh:
    ...:     fh.write(line.strip())

In [7]: calc_checksum('test')
Out[7]: '907737a0287cdf5569eaf857e22f7e8c test\n'

In [8]: read_checksum('version.md5')
Out[8]: '3fda9608a77b3809141f75a6069b4abd VERSION\n'

In [9]: with open('test','w') as fh:
    ...:     fh.write(line)

In [10]: calc_checksum('test')
Out[10]: '3fda9608a77b3809141f75a6069b4abd test\n'

```

autometa/config/databases.py
WiscEvan: The reason I'm only doing this for `nr.gz.md5` is because we have a reference md5 for `nr.gz` but we do not have a reference for `nr.dmnd.md5` a priori.

I think this is a subtle difference that may be more specific for issue-#71? I'm not sure

autometa/config/databases.py
WiscEvan: I think the user could add their databases within `checksums` as well as the `markers` section where the `option` should match in both.

WiscEvan: This is by default retrieving the md5 over http requests and not FTP. So far there is no mechanism in place to determine this based on the URL

autometa/config/databases.py
WiscEvan: I needed to push before being able to test. I will update with a comment when I have tested the functionality.

autometa/config/databases.py
WiscEvan: This is an oversight. 🐛

autometa/common/utilities.py
jason-c-kwan: OK, this makes sense.

autometa/config/databases.py
jason-c-kwan: Yes, perhaps. I just think that since the conversion to a .dmnd file takes a long time, it is important here to get a checksum at the end (only provided, of course, that the diamond command finished without error).

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+
+        """
+        hmms = (os.path.join(self.markers_dir, fn) for fn in os.listdir(self.markers_dir) if fn.endswith('.hmm'))
+        hmm_search_str = os.path.join('autometa/databases/markers/','*.h3*')
+        pressed_hmms = glob(hmm_search_str)
+        pressed_hmms = set(os.path.realpath(os.path.splitext(fp)[0]) for fp in pressed_hmms)
+        for fp in hmms:
+            if fp in pressed_hmms:
+                continue
+            hmmer.hmmpress(fp)
```
jason-c-kwan, line 348: Here it would make sense to write checksum files for the pressed databases when the command has finished successfully (or within hmmer.py, I guess). When these files are to be used, then we can check the checksum.

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
             if self.config.has_option('markers', option):
                 outfpath = self.config.get('markers', option)
             else:
                 outfname = os.path.basename(url)
                 outfpath = os.path.join(self.markers_dir, outfname)
+
             if self.dryrun:
                 logger.debug(f'UPDATE: (markers,{option}): {outfpath}')
                 self.config.set('markers', option, outfpath)
                 continue
```
jason-c-kwan, line 381: Instead of `continue`, shouldn't this be `return`? Otherwise if `self.dryrun == True`, then we still go ahead and download the markers.

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
-                hmmer.hmmpress(outfpath)
+        self.press_hmms()
+
+    def get_missing(self, section=None, validate=False):
+        """Get all missing database files in `options` from `sections`
+        in config.
+
+        If the database file already exists, an md5 checksum comparison will be
+        performed bewteen the current and file's remote md5 to ensure file integrity
+        prior to checking the respective file as valid.
```
jason-c-kwan, line 398: This function doesn't seem to do this though. If the file exists then it returns `True`. So we need to actually add the md5 comparison. Also, as written right now I think this will fail for the marker files at least because an md5 file is not made then they are downloaded (see other comment).

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+        sections = [section] if section else Databases.SECTIONS.keys()
+        invalid = {}
+        taxdump_checked = False
+        for section in sections:
+            for option in self.config.options(section):
+                if option not in Databases.SECTIONS.get(section):
+                    # Skip user added options not required by Autometa
+                    continue
+                # nodes.dmp, names.dmp and merged.dmp are all in taxdump.tar.gz
+                option = 'taxdump' if option in {'nodes','names','merged'} else option
```
jason-c-kwan, line 508: I made another comment elsewhere but here we are just checking that the parent taxdump file is OK but not the untarred component files.

autometa/config/databases.py
```diff
@@ -66,230 +97,504 @@ class Databases:
...
+            options = missing.get(section, [])
+            dispatcher[section](options)
+        else:
+            self.get_missing()
+            for section,options in missing.items():
+                dispatcher[section](options)
 
-    def update_missing(self):
-        """Download and format databases for all options in each section.
+    def compare_checksums(self, section=None, validate=False):
```
jason-c-kwan, line 476: I think we need a related function which looks at the checksum stored for a local file, then recalculates the file's checksum to see if they are the same. This could be used for files created by the pipeline itself to ensure the file is OK. If the md5 file is not there then that would signify that the step didn't finish properly.

jason-c-kwan: For example, this would be needed for the individual files that are untarred from taxdump. But it could also be applied to all other files made in other parts of the pipeline.

autometa/config/databases.py
```diff
@@ -298,46 +603,87 @@ def configure(self):
...
+        action='store_true', default=False)
+    parser.add_argument('--check-dependencies',
+        help='Check database dependencies are satisfied.',
+        action='store_true', default=False)
     parser.add_argument(
-        '--config', help='</path/to/input/database.config>', default=DEFAULT_FPATH)
-    parser.add_argument('--dryrun', help='Log configuration actions but do not perform them.',
-                        action='store_true', default=False)
+        '--no-checksum',
+        help='Do not perform checksum comparisons to validate integrity of database files',
```
jason-c-kwan, line 648: The description says this is to check the validity of the files. I think this is a little misleading for the user, because it could also trigger if the ftp site has newer databases even though the local files are fine. Perhaps modify the wording?

jason-c-kwan: If the related function mentioned in another comment is implemented, then this wording makes a bit more sense.

autometa/common/utilities.py
WiscEvan: Do you mean only when the function is instructed to remove the original file?

jason-c-kwan: No. After it makes the `out` file (successfully), it should call `write_checksum`. Probably should verify that there is a checksum file there for the original archive as well before using.

WiscEvan: 1. I'm not sure this is appropriate in the context of `gunzip`. It seems like this behavior should only be performed if specified.

2. Are we planning on hosting checksums for the database files that _do not_ have hosted checksums?

3. In relation to issue-#71, I was thinking of writing a general function as a decorator, so when we reach a step in the pipeline we would like Autometa to check, we can wrap the step with `@checkpoint` or some equivalent. Perhaps this would take as an argument (as well as the function) an output file path, such as `@utilities.checkpoint(out=outfpath)`. We would similarly require the function have an output file path argument to avoid confusion.

Example:

```python
@checkpoint
def some_func(infpath, out):
    """"This function takes as input an `infpath` and writes an `out`""""
    some_info = read(infpath)
    write(outfpath)
    return outfpath

# Since `@checkpoint` is specified
# `out` will be parsed from the function and checksummed
# then added to the checkpoints file for lookup later
```

I need to work on this implementation still, but for the future, just being able to place a `@checkpoint` decorator on a step in the pipeline that writes a file seems much simpler than writing out `checkpoint` functionality on a function-by-function basis.

autometa/config/databases.py
WiscEvan: Note: The taxdump.tar.gz download is fairly lightweight. I've also noticed this file gets updated multiple times daily.

autometa/config/databases.py
WiscEvan: The checksum check for the markers appears to be working

```python

In [1]: import requests

In [2]: outfpath = 'test.md5'

In [3]: with requests.Session() as session, open(outfpath, 'w') as fh:
   ...:     resp = session.get('https://raw.githubusercontent.com/WiscEvan/Autometa/databases/autometa/databases/markers/bacteria.single_copy.hmm.md5')
   ...:     if not resp.ok:
   ...:         raise 'ouch'
   ...:     fh.write(resp.text)
   ...:

In [4]: resp.text
Out[4]: 'eafdb5bd1447e814a4ee47ee11434ffe bacteria.single_copy.hmm\n'

In [5]: !cat test.md5
eafdb5bd1447e814a4ee47ee11434ffe bacteria.single_copy.hmm
```

autometa/config/databases.py
WiscEvan: Do you mean write a checksum for each of the four files that are created by `hmmpress`? 

e.g. 

1. `markers.hmm.h3p`
2. `markers.hmm.h3i`
3. `markers.hmm.h3m`
4. `markers.hmm.h3f`

autometa/config/databases.py
WiscEvan: This is looping through the `options`, so continuing will skip the downloading and amend the `config` with the option.

autometa/config/databases.py
WiscEvan: Both `FTP` and `requests` methods have been tested.

autometa/config/databases.py
WiscEvan: After thinking over the checksum writing this weekend. I agree with you on writing checksums after each successful download of each of the database files. As well as writing checksums for formatted files like `nr.dmnd` that take a long time to create. I'm not sure we need to checksum the formatted markers or the files within taxdump. Although maybe we write a database summary file and add in the checksums from the downloaded and formatted databases? So in this case we would calculate checksums for _all_ of the formatted files and write them to one database summary file. If any of the corresponding database checksums are updated, this would be reflected in the summary checksum. I'm thinking of logic similar to a Makefile here. This way we can update the formatted databases (`*.h3*`, `nr.dmnd`, `nodes.dmp`, `names.dmp`, `merged.dmp`) when any of their respective parent database files are updated while also ensuring file integrity later with lookup of the database summary during any Autometa run.

autometa/common/utilities.py
WiscEvan: > No. After it makes the `out` file (successfully), it should call `write_checksum`. Probably should verify that there is a checksum file there for the original archive as well before using.

Do you mean given `infpath`, look for `f'{infpath}.md5'`? If this does not exist, does this really tell us anything?

autometa/config/databases.py
WiscEvan: All files after they are downloaded are checksummed. I'll write checksums out for any of the formatted/extracted files.

See https://github.com/KwanLab/Autometa/pull/77#discussion_r426850961 for details

autometa/config/databases.py
WiscEvan: 1. `self.dryrun` is a user setting telling autometa _not_ to run the formatting/database downloading.
2. `no_update_db` (I've renamed to `do_not_upgrade`... Struggling to determine an appropriate variable name) checks that `nr.dmnd` exists _and_ the user does _not_ want to upgrade the database _even_ if the checksums do not match.
3. To toggle `checksums_match` to `True`, `os.path.exists(db_outfpath)` must also be `True`. So this ensures if the checksums match `nr.dmnd` will also exist.

autometa/config/databases.py
WiscEvan: I'm going to try returning an empty `dict` instead mixed dtypes (bool and dict) to try to make this more readable.

WiscEvan: Right now though,

if a user were to specify `validate = True`, then the `if validate: return True` block will _only_ be encountered if `if os.path.exists(fpath): continue` fails. Meaning there is a missing file, so return `True`. Otherwise, if `validate = True`, and no missing files are encountered, the function will return `False`.

autometa/config/databases.py
jason-c-kwan: Yes. If we have a convention that we always make a checksum (either as a separate file or in some special checksums file) after a file is created, then the pipeline can always check whether a checksum is available for a particular file, and we can be confident that the file is good if it's current checksum (i.e. recalculated) matches the one we have.

autometa/config/databases.py
jason-c-kwan: Yes, I think a single file for checksums would work. My argument for doing a checksum for every output file is that it is much more work devising all sorts of tests for the format of each file to check it is OK and not corrupted. Checksums are both (1) foolproof/accurate and (2) easy.

autometa/common/utilities.py
jason-c-kwan: Well, if you commit to always write a md5 file after any output step finishes without error, then it is an easy check later that you did that step, and you can then go ahead and check the file against the checksum. The two checks are:

1. md5 file exists - therefore step completed successfully at some point in the past
2. md5 checksum in the md5 file matches that which I just calculated with the file - therefore the file has not been messed with or corrupted since it was made.

autometa/config/databases.py
WiscEvan: This is being done now in `press_hmms`

autometa/common/utilities.py
WiscEvan: > Well, if you commit to always write a md5 file after any output step finishes without error, then it is an easy check later that you did that step, and you can then go ahead and check the file against the checksum. The two checks are:
> 
> 1. md5 file exists - therefore step completed successfully at some point in the past

We also discussed writing the checksums to one file at the completion of any output step here https://github.com/KwanLab/Autometa/issues/71#issuecomment-629479469. Do you think we should do both? There would be some redundant behavior. One file may reduce the clutter in the output directory as well as eliminate the need to perform multiple `file exists` checks.  Basically the commitment would be to write the checksum to the checkpoints file at each successful step rather than a written checksum at each step.

The analogous check from point 1. would be checking that the named output is in the checkpoints file. If it is, then we can retrieve the checksum value for comparison against the calculated checksum of the file (point 2).

jason-c-kwan: I don't think it matters conceptually, but there is a sort of "all eggs in one basket" problem with using one file - if that file gets corrupted then you are screwed. I mean, it won't be a big file so perhaps this is taking paranoia to a whole new level. Just thought I'd put that out there.

WiscEvan: Okay. I suppose it does not hurt to have an extra fail safe. Good point with the "all eggs in one basket", this makes sense. Currently the database files are being written with checksums. It sounds like for this, we pretty much have all of the functionality we need just from `write_checksum`, `calc_checksum` and `read_checksum`.

autometa/config/databases.py
jason-c-kwan: Just looked at the code changes in this function, and they look good. The one problem is that we are now checking that the `nr.dmnd.md5` file exists, but not that the checksum matches that of the file as it now exists on the disk.

autometa/config/databases.py
```diff
@@ -111,185 +232,388 @@ def format_nr(self):
...
+        into ncbi databases directory and update user config with extracted paths.
+
+        This only extracts nodes.dmp, names.dmp and merged.dmp from
+        taxdump.tar.gz if the files do not already exist. If `upgrade`
+        was originally supplied as `True` to the Databases instance, then the
+        previous files will be replaced by the new taxdump files.
 
-        Extracts nodes.dmp, names.dmp and merged.dmp from taxdump.tar.gz
+        After successful extraction of the files, a checksum will be written of the
+        archive for future checking and then the archive will be removed to save user disk space.
```
jason-c-kwan, line 277: I think this is wrong - I couldn't see where the .tar file is deleted in this function, so the comment should be changed. For what it is worth, I think we should leave the parent .tar.gz file on the disk.

autometa/config/databases.py
jason-c-kwan: I saw the addition of `write_checksum` - this is not quite right yet, because if the download was corrupted then the checksum will probably match but be wrong. So we should additionally check that the written checksum file matches the remote checksum at this point.

autometa/config/databases.py
```diff
@@ -111,185 +232,388 @@ def format_nr(self):
...
+                # nodes.dmp, names.dmp and merged.dmp are all in taxdump.tar.gz
+                option = 'taxdump' if option in {
+                    'nodes', 'names', 'merged'} else option
                 fpath = self.config.get(section, option)
-                if os.path.exists(fpath) and os.stat(fpath).st_size >= 0:
-                    # TODO: [Checkpoint validation]
-                    logger.debug(f'({section},{option}): {fpath}')
+                fpath_md5 = f'{fpath}.md5'
+                # We can not checksum a file that does not exist.
+                if not os.path.exists(fpath) and not os.path.exists(fpath_md5):
```
jason-c-kwan, line 536: If `fpath` exists but `fpath_md5` doesn't exist, doesn't that suggest some problem occurred during download? More to the point, should we do something in this case rather than skipping with `continue`?

WiscEvan:
Regarding https://github.com/KwanLab/Autometa/pull/77#discussion_r429446790. This will `gunzip` anything passed to it. So whether it is called on a downloaded file or otherwise is dependent on how the function is used.

autometa/config/databases.py
```diff
@@ -111,185 +232,388 @@ def format_nr(self):
...
+            current_checksum = read_checksum(db_infpath_md5)
+            current_hash, __ = current_checksum.split()
+            host = self.config.get('ncbi', 'host')
+            remote_checksum = self.get_remote_checksum(host, 'nr')
+            remote_hash, __ = remote_checksum.split()
+            if current_hash == remote_hash:
+                logger.debug(f'nr checksums match, skipping...')
+                checksums_match = True
+        dmnd_md5 = f'{db_outfpath}.md5'
+        do_not_upgrade = os.path.exists(db_outfpath) and not self.upgrade
```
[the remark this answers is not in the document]
WiscEvan: The `checksums_match` variable is checked with `do_not_upgrade` in line 250. This way, the user can specify to keep the database _even_ if the checksums _do not_ match, i.e. they do not mind a database that is a few weeks out-of-date. `do_not_upgrade` will only evaluate to True if `nr.dmnd.` exists and the user specifies _not_ to upgrade the database to the most up-to-date version.

autometa/config/databases.py
WiscEvan: I've removed the lines that discard `taxdump.tar.gz`. This routine was at the end of the function. Lines 307-309

autometa/config/databases.py
WiscEvan: This only skips if _both_ the `fpath` and the `fpath_md5` do not exist

autometa/config/databases.py
WiscEvan: This has been addressed in the new commits

autometa/config/databases.py
WiscEvan: I'm not sure how we would do these comparisons without having the remote checksum available. Otherwise, we can perform the checksum comparisons like I am doing with the database files. A general checksum comparison for files in the pipeline is likely to be a bit different because there will is not a remote checksum. `compare_checksums` here is specific to databases. Another checksum comparison function can be added to utilities to aid in comparing pipeline generated files.

autometa/config/databases.py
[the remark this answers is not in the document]
jason-c-kwan: My point was that if `f'{db_outfpath}.md5'` does not match the current checksum of `db_outfpath` then it should be remade, and perhaps that should be checked here if it would not be caught later?

autometa/config/databases.py
```diff
@@ -367,26 +390,26 @@ def press_hmms(self):
...
+        hmm_search_str = os.path.join(self.markers_dir, "*.h3?")
         # First search for pressed hmms to remove from list to hmmpress
-        pressed_hmms = set(
+        pressed_hmms = {
             os.path.realpath(os.path.splitext(fp)[0])
             for fp in glob(hmm_search_str)
-            if not fp.endswith('.md5')
-        )
+            if not fp.endswith(".md5")
+        }
```
jason-c-kwan: I'm not sure I understand what is going on here. Instead of being cast as a `set`, the filepaths not ending in `'.md5'` are cast as a `dict`? Why is this change needed, and does it just become a dictionary with keys and no values?

WiscEvan: `pressed_hmms` is a set. This is an example of a set comprehension. Instead of constructing a list then casting the list as a set, a set is simply constructed.

The syntax for dictionary or set comprehensions are similar to list comprehensions.

```python
list_comprehension = [elem for elem in elems]
# Notice curly braces
set_comprehension = {elem for elem in elems}
# Notice curly braces _and_ the colon to separe the key:value pairs
dict_comprehension = {i: elem for i, elem in enumerate(elems)}

```

jason-c-kwan: Yeah, sorry. For whatever reason I have not seen that before.

autometa/config/databases.py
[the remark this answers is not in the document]
jason-c-kwan: Can you explain the reasoning behind lines 261-266? It looks like if the md5 file exists, then it recalculates the md5 from the diamond file. But then if they don't match the old one is re-written?? Wouldn't this cover up a corrupted file?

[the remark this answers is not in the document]
jason-c-kwan: This I think is the only outstanding issue that has not been resolved yet.

jason-c-kwan:
This was probably lost in the weeds above, but I had an outstanding comment on `databases.py`, reproduced here:

Can you explain the reasoning behind lines 261-266? It looks like if the md5 file exists, then it recalculates the md5 from the diamond file. But then if they don't match the old one is re-written?? Wouldn't this cover up a corrupted file?

WiscEvan:
> This was probably lost in the weeds above, but I had an outstanding comment on `databases.py`, reproduced here:
> 
> Can you explain the reasoning behind lines 261-266? It looks like if the md5 file exists, then it recalculates the md5 from the diamond file. But then if they don't match the old one is re-written?? Wouldn't this cover up a corrupted file?

I added this after you mentioned wanting to overwrite the existing file if the current database checksum did not match. See https://github.com/KwanLab/Autometa/pull/77#discussion_r440831802

> My point was that if `f'{db_outfpath}.md5'` does not match the current checksum of `db_outfpath` then it should be remade, and perhaps that should be checked here if it would not be caught later?

I think I've resolved this. I interpretted the comment above differently than intended I think. I think now we are checking all of the bases.

autometa/config/databases.py
```diff
@@ -245,31 +245,37 @@ def format_nr(self):
...
         dmnd_md5 = f"{db_outfpath}.md5"
-        if os.path.exists(dmnd_md5) and os.path.exists(db_outfpath):
-            dmnd_md5_hash = read_checksum(dmnd_md5).split()[0]
-            db_outfpath_hash = calc_checksum(db_outfpath).split()[0]
-            if dmnd_md5_hash != db_outfpath_hash:
-                write_checksum(db_outfpath, dmnd_md5)
+        if os.path.exists(dmnd_md5) and db_outfpath_exists:
+            dmnd_md5_hash, __ = read_checksum(dmnd_md5).split()
+            if db_outfpath_hash == dmnd_md5_hash:
+                checksums_match = True
```
jason-c-kwan, line 269: `checksums_match` should be set to `False` before this block, because it is likely to have been set to `True` further up. Therefore, in the failure case where `db_outfpath_hash != dmnd_md5_hash`, it will still be `True`.

Issue #42, opened 2020-04-12
[dev branch] Issued with autometa/taxonomy/majority_vote.py


jason-c-kwan:
This file contains the functions which carry out the majority vote algorithm (basically this takes the lowest common ancestor/LCA information for each ORF from each contig, and converts this information into a single taxid for each contig). The algorithm is the same as in Autometa 1.0, and I didn't find any problems with the actual logic of it. Other issues with the script are outlined below:

1. [x] The docstring of `majority_vote` is not very descriptive for all the arguments.

2. [x] In the `majority_vote` function, we call the `lca` and `voted_taxids` functions, and use `dict.get` expressions to bring up relevant arguments. However, in the `lca` call, the default for `verbose` is `True`, whereas in the `voted_taxids` call, the default for `verbose` is `False`. These should both probably match the command line default for `verbose`, which is `False`.

https://github.com/KwanLab/Autometa/blob/59941204e1ebb59a04da01b9bbd46c7fd50d3e7b/autometa/taxonomy/majority_vote.py#L266-L272

https://github.com/KwanLab/Autometa/blob/59941204e1ebb59a04da01b9bbd46c7fd50d3e7b/autometa/taxonomy/majority_vote.py#L287-L291

3. [x] In the `rank_taxids` function, the docstring is not very descriptive about what sort of data structure the argument `ctg_lcas` should be.

4. [x] In the `rank_taxids` function, we initialize an `NCBI` object in order to access various aspects of the taxonomy tree. However, `rank_taxids` is called by `majority_vote`, and in the latter function, we already created an `LCA` object. Because the `LCA` class inherits from `NCBI`, the initialization of each involves parsing through the entire NCBI taxonomy and storing it in memory. So, as written currently we are wasting time and memory unnecessarily (i.e. we parse the NCBI taxonomy twice into two separate copies in memory). It would be more efficient to pass the `LCA` object to `rank_taxids`. Other functions in the script expect an NCBI object, and would be fine with the LCA object instead.

[...]

PR #81, opened 2020-05-14
Update majority_vote

WiscEvan:
* fixes issue-#42
* :memo: Add/Update docstrings in functions.
* :art::racehorse: Change ncbi_dir parameter to ncbi in rank_taxids to take ncbi instance rather than instantiating in function.
* :art::memo: Rename ctg_lcas parameter to rank_counts where dict corresponds to only one contig.

### DIFF
--- a/autometa/taxonomy/majority_vote.py
+++ b/autometa/taxonomy/majority_vote.py
@@ -221,15 +221,15 @@ def majority_vote(fasta, ncbi_dir, outdir, votes_fname, lca_fname=None, **kwargs
     Parameters
     ----------
     fasta : str
-        </path/to/prot/orfs.faa>
+        Path to ORFs fasta containing amino-acid sequences to be annotated.
     ncbi_dir : str
-        </path/to/ncbi/dir>
+        Path to NCBI databases directory.
     outdir : str
-        <path/to/output/dir>
+        Path to output directory to store intermediate results.
     votes_fname : str
-        <votes.tsv filename>
+        Output filename of assigned taxids. Note: Will be written to `outdir`.
     lca_fname : str, optional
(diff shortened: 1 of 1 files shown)


autometa/taxonomy/majority_vote.py
```diff
@@ -230,7 +215,6 @@ def rank_taxids(ctg_lcas, ncbi_dir, verbose=False):
         top_taxids[contig] = acceptedTaxid
     return top_taxids
 
-
 def majority_vote(fasta, ncbi_dir, outdir, votes_fname, lca_fname=None, **kwargs):
     """Wrapper for modified majority voting algorithm from Autometa 1.0
```
jason-c-kwan, line 220: See first point in issue #42. More detail about the arguments would be nice, for example presumably `lca_fname` will be the filename written in `outdir`?

autometa/taxonomy/majority_vote.py
```diff
@@ -315,40 +294,46 @@ def write_votes(results, outfpath):
...
         raise FileExistsError(outfpath)
-    with open(outfpath, 'w') as fh:
-        fh.write(outlines)
+    lines = 'contig\ttaxid\n'
+    fh = open(outfpath, 'w')
+    count = 0
+    for contig, taxid in results.items():
+        if count >= 60000:
+            fh.write(lines)
+            lines = ''
```
jason-c-kwan, line 309: Count should be reset to zero after this.

autometa/taxonomy/majority_vote.py
jason-c-kwan: Can probably copy/paste from the help messages in `main`.

PR #84, opened 2020-05-19
Recursive DBSCAN

### LINKED ISSUE #52: [dev branch] Issues with autometa/binning/recursive_dbscan.py
I hope it is OK to not use one of the issue templates, but I couldn't see "code review" there. This script contains functions required for clustering contigs and recursively searching for bins with the highest completeness and purity. I noticed the following issues as part of my code review.

1. [x] In the `recursive_dbscan` function, during each iteration the uses a different value for `eps`, we calculate the median completeness of the bins that pass the completeness and purity cutoffs. As stated in the NAR article, the `eps` value which gives the greatest median completeness is considered the best. The code that calculates the median is shown below:

https://github.com/KwanLab/Autometa/blob/3e8c7718dc63e5a06654a899506535ae8bd6219e/autometa/binning/recursive_dbscan.py#L103

However, the problem here is that the dataframe `df` has rows for each contig, and a column for 'completeness' (i.e. each contig in a particular cluster will have the same value for completeness under this column heading). So, what that means is that the above code will take a completeness value for each contig, and calculate the median from that. This is not what is intended - in the last version what was done was a completeness value for each individual cluster was used to calculate median. The way it is done above, there will be biases depending on the number of contigs a cluster has. I would suggest something like the following:

```python
import statistics
cluster_completeness = dict() # Keyed by cluster
clusters = dict(list(df.groupby('cluster'))).keys()
for index, row in df.iterrows():
    cluster = row['cluster']
    if cluster in cluster_completeness:
        continue
    else:
        completeness = row['completeness']
        cluster_completeness[cluster] = completeness   
median_completeness = statistics.median(list(cluster_completeness.values()))
```
Note: I am not a Pandas expert, so there could be a better way of doing the above.

2. [x] In the `run_dbscan` function, there is no description of the `df` and `eps` arguments in the docstring.

3. [x] In the `run_dbscan` function, there is an initial check to see of `n_samples` (i.e. the number of rows) is 1. If it is, then it is kind of pointless to do clustering on it, and it looks like the intended behavior is that the cluster for the one contig be recorded as `pd.NA`, before the dataframe is returned:

https://github.com/KwanLab/Autometa/blob/3e8c7718dc63e5a06654a899506535ae8bd6219e/autometa/binning/recursive_dbscan.py#L172-L177

There is a potential problem here - immediately after this check we drop some columns from the dataframe, which by default are `['cluster', 'purity', 'completeness']`. I think that this drop columns part should be before the 1 row check, because the returned 1 row dataframe might have old cluster, purity and completeness values which will be left there (I actually don't know what will happen during the merge statement if there is already a column called 'cluster').

4. [x] There is a bit of confusion in the `run_dbscan` function about the embedded dimensions. In the def line, it says that the default value of `usecols` is `['x', 'y']` (i.e. DBSCAN on two embedded dimensions, excluding coverage). However, in the docstring, it says that the default is `['x', 'y', 'coverage']` if `coverage` is present in the dataframe, otherwise it is `['x', 'y', 'z']`. Leaving aside the problem that we aren't given the option of using three embedded dimensions without coverage, it turns out that in the function body `usecols` is never used. Instead, we first add `['x', 'y']` to `cols`, then if `coverage` is in the dataframe we use that, otherwise we add `z`. One problem with this behavior, is if `coverage` is not present in the table, then it doesn't check whether `z` is there, so there may be an error in that circumstance. I would suggest that we implement the default behavior outlined in the docstring properly, but have a check that `z` exists. Perhaps there should also be an option in case the user wants to just use `x` and `y`.

5. [x] In the `add_metrics` function, there is no description of the argument `df` in the docstring.

6. [x] In the `add_metrics` function, there is a `domain` argument which tells the function how many markers to expect in a cluster. We translate this to an expected total like this:

https://github.com/KwanLab/Autometa/blob/ddc2bf951d6c4a8c29e9e2301e15a762327b0267/autometa/binning/recursive_dbscan.py#L213-L214

The `dict.get` method here searches for the `domain` in `marker_sets`, and if it doesn't find the key then it returns `139`. Instead, there should probably be an error thrown with a message saying that the domain is not recognized. This should aid troubleshooting (for example - a script could contain archaea misspelled or something).

7. [x] In the `binning` function, if the `master` dataframe doesn't have any marker information in it, then we throw a `RecursiveDBSCANError`. Since we have also implemented HDBSCAN it should probably be something like `RecursiveBinningError`. Come to think of it, the script name `recursive_dbscan.py` is also a bit specific.

WiscEvan:
* resolves #52 
* :memo: Update docstrings
* :bug::racehorse:Update median completeness calculation to not incorporate all contigs but rather just cluster values.
* :art: Change `RecursiveDBSCANError` exception to `BinningError`
* :bug: Update `metabin` import in `user.py` from `mag.py`
* :art::fire: Remove default 'z' column in `run_dbscan` function and `usecols` parameters.
* :art: `add_metrics` func now returns 2-tuple with cluster_metrics dataframe for median completeness lookup.
* :art: Change default domain marker count lookup in `add_metrics` to raise error if domain does not match.

### DIFF
--- a/autometa/binning/recursive_dbscan.py
+++ b/autometa/binning/recursive_dbscan.py
@@ -36,19 +36,127 @@ from hdbscan import HDBSCAN
 
 from autometa.common.markers import Markers
 from autometa.common import kmers
+
 # TODO: This should be from autometa.common.kmers import Kmers
 # So later we can simply/and more clearly do Kmers.load(kmers_fpath).embed(method)
 from autometa.common.exceptions import BinningError
 from autometa.taxonomy.ncbi import NCBI
 
-pd.set_option('mode.chained_assignment', None)
+pd.set_option("mode.chained_assignment", None)
 
 
 logger = logging.getLogger(__name__)
 
 
-def recursive_dbscan(table, markers_df, domain, completeness_cutoff,
-                     purity_cutoff, verbose=False, method='DBSCAN'):
+def add_metrics(df, markers_df, domain="bacteria"):
+    """Adds the completeness and purity metrics to each respective contig in df.
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        index='contig' cols=['x','y','coverage','cluster']
+    markers_df : pd.DataFrame
+        wide format, i.e. index=contig cols=[marker,marker,...]
+    domain : str, optional
+        Kingdom to determine metrics (the default is 'bacteria').
+        choices=['bacteria','archaea']
+
+    Returns
+    -------
+    2-tuple
+        `df` with added cols=['completeness', 'purity']
+        pd.DataFrame(index=clusters,  cols=['completeness', 'purity'])
+
+    Raises
+    -------
+    KeyError
+        `domain` is not "bacteria" or "archaea"
+
+    """
+    domain = domain.lower()
+    marker_sets = {"bacteria": 139.0, "archaea": 162.0}
+    if domain not in marker_sets:
+        raise KeyError(f"{domain} is not bacteria or archaea!")
+    expected_number = marker_sets[domain]
+    clusters = dict(list(df.groupby("cluster")))
+    metrics = {"purity": {}, "completeness": {}}
+    for cluster, dff in clusters.items():
+        contigs = dff.index.tolist()
+        summed_markers = markers_df[markers_df.index.isin(contigs)].sum()
+        is_present = summed_markers >= 1
+        is_single_copy = summed_markers == 1
+        nunique_markers = summed_markers[is_present].index.nunique()
+        num_single_copy_markers = summed_markers[is_single_copy].index.nunique()
+        completeness = nunique_markers / expected_number * 100
+        # Protect from divide by zero
+        if nunique_markers == 0:
+            purity = pd.NA
+        else:
+            purity = num_single_copy_markers / nunique_markers * 100
+        metrics["completeness"].update({cluster: completeness})
+        metrics["purity"].update({cluster: purity})
+    metrics_df = pd.DataFrame(metrics, index=clusters.keys())
+    merged_df = pd.merge(df, metrics_df, left_on="cluster", right_index=True)
+    return merged_df, metrics_df
+
+
+def run_dbscan(df, eps, dropcols=["cluster", "purity", "completeness"]):
+    """Run clustering on `df` at provided `eps`.
+
+    Notes
+    -----
+
+        * documentation for sklearn `DBSCAN <https://scikit-learn.org/stable/modules/generated/sklearn.cluster.DBSCAN.html>`_
+        * documentation for `HDBSCAN <https://hdbscan.readthedocs.io/en/latest/index.html>`_
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Contigs with embedded k-mer frequencies as ['x','y'] columns and optionally 'coverage' column
+    eps : float
+        The maximum distance between two samples for one to be considered
+        as in the neighborhood of the other. This is not a maximum bound
+        on the distances of points within a cluster. This is the most
+        important DBSCAN parameter to choose appropriately for your data set
+        and distance function. See `DBSCAN docs <https://scikit-learn.org/stable/modules/generated/sklearn.cluster.DBSCAN.html>`_ for more details.
+    dropcols : list, optional
+        Drop columns in list from `df`
+        (the default is ['cluster','purity','completeness']).
+
+    Returns
+    -------
+    pd.DataFrame
+        `df` with 'cluster' column added
+
+    Raises
+    -------
+    ValueError
+        sets `usecols` and `dropcols` may not share elements
+
+    """
+    for col in dropcols:
+        if col in df.columns:
+            df.drop(columns=col, inplace=True)
+    n_samples = df.shape[0]
+    if n_samples == 1:
+        clusters = pd.Series([pd.NA], index=df.index, name="cluster")
+        return pd.merge(df, clusters, how="left", left_index=True, right_index=True)
+    cols = ["x", "y"]
+    if "coverage" in df.columns:
+        cols.append("coverage")
+    if np.any(df.isnull()):
+        raise BinningError(
+            f"df is missing {df.isnull().sum().sum()} kmer/coverage annotations"
+        )
+    X = df.loc[:, cols].to_numpy()
+    clusterer = DBSCAN(eps=eps, min_samples=1, n_jobs=-1).fit(X)
+    clusters = pd.Series(clusterer.labels_, index=df.index, name="cluster")
+    return pd.merge(df, clusters, how="left", left_index=True, right_index=True)
+
+
+def recursive_dbscan(
+    table, markers_df, domain, completeness_cutoff, purity_cutoff, verbose=False,
+):
     """Carry out DBSCAN, starting at eps=0.3 and continuing until there is just one
     group.
 
@@ -76,9 +184,6 @@ def recursive_dbscan(table, markers_df, domain, completeness_cutoff,
         `purity_cutoff` threshold to retain cluster (the default is 90.0).
     verbose : bool
         log stats for each recursive_dbscan clustering iteration.
-    method : str
-        clustering `method` to perform. (the default is 'DBSCAN').
-        Choices=['DBSCAN','HDBSCAN'] NOTE: 'HDBSCAN' is still under development.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


WiscEvan:
Finally trying out `vscode` with auto-formatting. Not entirely satisfied with it to be honest.

jason-c-kwan:
@WiscEvan Could we customize a template or something? I've still not used it yet.

WiscEvan:
> @WiscEvan Could we customize a template or something? I've still not used it yet.

Probably a better question for @chanana, @Sidduppal or @kaw97 I think all three have been using this far longer than I have.

WiscEvan:
> Finally trying out `vscode` with auto-formatting. Not entirely satisfied with it to be honest.

I've stumbled on [black](https://black.readthedocs.io/en/stable/) which will format consistently given any file and any number of formatting runs. I think this will be easiest moving forward so we no longer will need to worry about consistency when formatting each script. black can just handle everything.

autometa/binning/recursive_dbscan.py
```diff
@@ -91,16 +93,18 @@ def recursive_dbscan(table, markers_df, domain, completeness_cutoff, purity_cuto
...
-        df = add_metrics(df=binned_df, markers_df=markers_df, domain=domain)
-        completess_filter = df['completeness'] >= completeness_cutoff
-        purity_filter = df['purity'] >= purity_cutoff
-        median_completeness = df[completess_filter & purity_filter]['completeness'].median()
+        df, metrics_df = add_metrics(
+            df=binned_df, markers_df=markers_df, domain=domain)
+        completeness_filter = metrics_df['completeness'] >= completeness_cutoff
+        purity_filter = metrics_df['purity'] >= purity_cutoff
+        median_completeness = metrics_df[completeness_filter &
+                                         purity_filter]['completeness'].median()
```
jason-c-kwan, line 107: This doesn't use groupby methods like in your comment on issue #52. So will this calculate the median completeness of each bin?

WiscEvan: Yes,

```python
# metrics_df is indexed by cluster.
>>>metrics_df[completeness_filter & purity_filter].head()
       purity  completeness
0     96.4286     20.143885
1         100     27.338129
5     96.3636     39.568345
6         100     22.302158
7     97.7528     64.028777

>>>metrics_df[completeness_filter & purity_filter].median()
purity          100.000000
completeness     27.338129
dtype: float64

>>>metrics_df[completeness_filter & purity_filter]["completeness"].median()
27.33812949640288

```

WiscEvan:
I am making a few more edits here for an implementation of a `recursive_hdbscan` function.

jason-c-kwan:
So I should hold off on reviewing?

WiscEvan:
> So I should hold off on reviewing?

Yes, I am testing the HDBSCAN implementation now. I will be pushing the commit later today

autometa/binning/recursive_dbscan.py
```diff
@@ -92,73 +197,72 @@ def recursive_dbscan(table, markers_df, domain, completeness_cutoff,
...
-        n_clusters = df['cluster'].nunique()
+        n_clusters = df["cluster"].nunique()
         if n_clusters in clustering_rounds:
             clustering_rounds[n_clusters] += 1
         else:
             clustering_rounds[n_clusters] = 1
         # We speed this up if we are getting a lot of tables with the same number of clusters
         if clustering_rounds[n_clusters] > 10:
-            step *= 10
+            eps *= 10
```
jason-c-kwan, line 223: This changes behavior. Now the step change will depend on the current value of eps. Is this intended? Before we would have `step = 0.1`, and then if we had more than 10 rounds with the same number of clusters, it would change to `1.0`.

autometa/binning/recursive_dbscan.py
```diff
@@ -92,73 +197,72 @@ def recursive_dbscan(table, markers_df, domain, completeness_cutoff,
...
             break
         if verbose:
             logger.debug(
-                f'EPS: {eps:3.2f} Clusters: {n_clusters:,}'
-                f' Completeness: Median={median_completeness:4.2f} Best={best_median:4.2f}'
+                f"EPS: {eps:3.2f} Clusters: {n_clusters:,}"
+                f" Completeness: Median={median_completeness:4.2f} Best={best_median:4.2f}"
             )
-        eps += step
+        eps += 0.1
```
jason-c-kwan, line 233: See above - this doesn't do the same thing as before.

autometa/binning/recursive_dbscan.py
```diff
@@ -92,73 +197,72 @@ def recursive_dbscan(table, markers_df, domain, completeness_cutoff,
...
+        f"clustered: {len(complete_and_pure_df)} unclustered: {len(unclustered_df)}"
+    )
     return complete_and_pure_df, unclustered_df
 
 
-def run_dbscan(df, eps, dropcols=['cluster', 'purity', 'completeness'],
-               method='dbscan'):
-    """Run clustering on `df` via `method` at provided `eps`.
+def run_hdbscan(df, min_cluster_size, dropcols=["cluster", "purity", "completeness"]):
+    """Run clustering on `df` at provided `eps`.
```
jason-c-kwan, line 252: This should be changed as hdbcan doesn't have an eps parameter

autometa/binning/recursive_dbscan.py
```diff
@@ -92,73 +197,72 @@ def recursive_dbscan(table, markers_df, domain, completeness_cutoff,
...
 
     Parameters
     ----------
     df : pd.DataFrame
         Contigs with embedded k-mer frequencies as ['x','y'] columns and optionally 'coverage' column
-    eps : float
+    min_cluster_size : int
         The maximum distance between two samples for one to be considered
         as in the neighborhood of the other. This is not a maximum bound
         on the distances of points within a cluster. This is the most
```
jason-c-kwan: Isn't this the minimum number of points that can be in a cluster?

jason-c-kwan: Also a few lines below (sorry - this is grey on my view so I can't select), there is mention of DBSCAN, so perhaps we have the wrong description here.

autometa/binning/recursive_dbscan.py
```diff
@@ -180,95 +281,131 @@ def run_dbscan(df, eps, dropcols=['cluster', 'purity', 'completeness'],
...
-        is_single_copy = summed_markers == 1
-        nunique_markers = summed_markers[is_present].index.nunique()
-        num_single_copy_markers = summed_markers[is_single_copy].index.nunique(
-        )
-        completeness = nunique_markers / expected_number * 100
-        # Protect from divide by zero
-        if nunique_markers == 0:
-            purity = pd.NA
+
+    min_cluster_size = 2
```
jason-c-kwan, line 348: Can we start with `min_cluster_size = 1`? This would be equivalent to the `min_samples = 1` parameter we use for DBSCAN. The reason why I chose that is that there can be single contigs which legitimately are also a whole bin. For example AB1_lowGC that we found in the first *Bugula* metagenome.

jason-c-kwan: You should probably test this on AB1, because I am wondering whether this method of sweeping the parameter with HDBSCAN would mean that AB1_lowGC is always grouped with other things.

autometa/binning/recursive_dbscan.py
WiscEvan: Yikes, missed updating this. 🤦‍♂️

autometa/binning/recursive_dbscan.py
WiscEvan: This results in a ValueError from hdbscan:


<details open><summary><code>(autometa) ➜  Autometa git:(WiscEvan/issue52) ✗ python -m autometa.binning.recursive_dbscan --clustering-method hdbscan --taxonomy tests/dtc_data/taxonomy.tsv --embedded-kmers tests/dtc_data/kmers.bhsne.tsv tests/dtc_data/kmers.tsv tests/dtc_data/coverages.tsv tests/dtc_data/bacteria.markers.tsv tests/dtc_data/binning.tsv</code></summary><p>


```
06/18/2020 01:04:43 PM : autometa.common.kmers : DEBUG : k-mers frequency embedding already exists tests/dtc_data/kmers.bhsne.tsv
06/18/2020 01:04:43 PM : root : DEBUG : master_df shape: (2829, 11)
06/18/2020 01:04:43 PM : __main__ : INFO : Examining superkingdom: 1 unique taxa (2,829 contigs)
06/18/2020 01:04:43 PM : __main__ : DEBUG : Examining taxonomy: superkingdom : bacteria : (2829, 11)
06/18/2020 01:04:43 PM : __main__ : INFO : clustering with recursive_hdbscan
Traceback (most recent call last):
  File "/Users/rees/miniconda3/envs/autometa/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/rees/miniconda3/envs/autometa/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/rees/Autometa/autometa/binning/recursive_dbscan.py", line 710, in <module>
    main()
  File "/Users/rees/Autometa/autometa/binning/recursive_dbscan.py", line 701, in main
    verbose=args.verbose,
  File "/Users/rees/Autometa/autometa/binning/recursive_dbscan.py", line 594, in binning
    verbose=verbose,
  File "/Users/rees/Autometa/autometa/binning/recursive_dbscan.py", line 446, in get_clusters
    master_df, markers_df, domain, completeness, purity, verbose=verbose,
  File "/Users/rees/Autometa/autometa/binning/recursive_dbscan.py", line 352, in recursive_hdbscan
    binned_df = run_hdbscan(table, min_cluster_size)
  File "/Users/rees/Autometa/autometa/binning/recursive_dbscan.py", line 305, in run_hdbscan
    ).fit(X)
  File "/Users/rees/miniconda3/envs/autometa/lib/python3.7/site-packages/hdbscan/hdbscan_.py", line 919, in fit
    self._min_spanning_tree) = hdbscan(X, **kwargs)
  File "/Users/rees/miniconda3/envs/autometa/lib/python3.7/site-packages/hdbscan/hdbscan_.py", line 494, in hdbscan
    raise ValueError('Min cluster size must be greater than one')
ValueError: Min cluster size must be greater than one
```
</p></details>

Issue #36, opened 2020-04-06
[Dev branch] Issues with autometa/common/external/diamond.py


jason-c-kwan:
This file contains functions and the DiamondResult class used to running and processing Diamond blast results. Here are the issues that I noticed during code review.

1. [x] In the script, there is an argument/variable called `top_pct` or a variant of this name. This is a float value between 0 and 1 (although it can't exactly be zero), used to determine if the bitscore of hits is above a threshold. For example, if it is 0.9 then only bitscores >= 0.9 * the top bitscore are accepted. The problem here is not one of programmatic logic, but rather readability for contributors. `top_pct` looks like "top percentage", which would suggest, for example, that for a value of 0.9, 90% will be kept. However, the opposite is true in this case - only 10% will be kept. This is compounded by the help message in the `if __name__ == '__main__':` block, which describes this argument as "top percentage of hits to retrieve". So what I am saying is that the variable name and documentation is confusing. Perhaps it could be called "bitscore_tolerance" or something? Accompanied by a change in documentation and inclusion of explanatory comments.

2. [x] `makedatabase` function. I believe that database processing can be parallelized in Diamond, so it doesn't make sense here to have the default number of processors `1`. (Note: the `blast` function seems to have a default cpu value of zero, so you should double check whether Diamond interprets this as using all available cpus). This function also does not have a docstring.

3. [x]  `blast` function. This function carries out an actual Diamond blast search, and one of its arguments is `blast_type`, which is supposed to only have the values `blastp` or `blastx` (which I think are the only ones that Diamond can do). However, later on when the command is being constructed, `blastp` is hardcoded. So This should be changed to allow the function's `blast_type` argument to be used.

https://github.com/KwanLab/Autometa/blob/59941204e1ebb59a04da01b9bbd46c7fd50d3e7b/autometa/common/external/diamond.py#L192-L210

4. [ ] In the function `add_taxids` the docstring mentions that there is a todo task to make this function handle instances where taxids found in the prot.accession2taxid.gz file are not present in the nodes.dmp file (for instance by checking the merged.dmp file). This could be tackled if you would like to do it, and would probably require the involvement of the NCBI class. However, the main thing here is that the docstring should probably tell us that at present there is no handling of this, and taxids might be returned as `None` if there is a problem.

5. [x] The `add_taxids` function allows the accessing of either an uncompressed prot.accession2taxid file or the gzipped prot.accession2taxid.gz version. When the filehandle is opened, there is a choice depending on the filename:

[...]

PR #87, opened 2020-06-04
diamond.py

Sidduppal:
-  :art: Renames top_pct to bitscr_thresh which translates to bitscore threshold
-  :racehorse:  Uses all the available cpus as default 
-  :bug: Removes hardcoding of BLASTP
-  :art: Enables the searching of merged.dmp in case the taxid is not found in nodes.dmp
-  :art: :racehorse: gunzipped database opens with 'rt' mode
-  :memo:  Formats docstrings
- :bug: Fixes the code for `__sub__`


autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+                "sstart": int(sstart),
+                "send": int(send),
+                "evalue": float(evalue),
+                "bitscore": float(bitscore),
+            }
+
+        Returns
+        -------
+        Dict
+            {qseqid: {DiamondResult}, ...}
```
WiscEvan, line 221: Same comments as above. `other_hit` is not a `dict`, a `dict` is not returned and the parameter string is wrong

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+        """
+        assert self == other_hit, f"qseqids do not match! {self} & {other_hit}"
+        self.sseqids.pop(list(other_hit.sseqids.keys())[0])
```
WiscEvan, line 224: This is only removing the first `sseqid` from `self.sseqids`. This may also fail if the `other_hit.sseqids` are not in `self.sseqids`. You should be able to rework this so it will not raise a KeyError when trying to remove `other_hits.sseqids` that are not in the `self.sseqids`. This also needs to be fixed to follow the intended behavior

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+    Creates a database against which the query sequence would be blasted
+
+    Parameters
+    ----------
+    fasta : str
+        Path to fasta file whose database needs to be made
+        e.g. '<path/to/fasta/file>'
+    database : str
+        Path to the output  nr database file
+        e.g. '<path/to/database/file>'
```
WiscEvan, line 251: These parameters are somewhat connected. You are referring to `nr` in the database parameter. But this function can handle _any_ fasta file and write out a respective diamond formatted database file from the provided `fasta`. I'm not sure `nr` should be mentioned here. The `database` parameter description should be more clear.

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+    nproc : int, optional
+        Number of processors to be used. By default uses all the processors of the system 
+
+    Returns
+    -------
+    str
+        Path to Diamond database
```
WiscEvan, line 258: This should probably read "Path to diamond formatted database"

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+
+    Raises
+    ------
+    CalledProcessError
+        Failed to create Diamond database
```
WiscEvan, line 263:
```suggestion
        Failed to create diamond formatted database
```

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+        Number of processors to be used. By default uses all the processors of the system 
+
+    Returns
+    -------
+    str
+        Path to Diamond database
+
+    Raises
+    ------
+    CalledProcessError
```
WiscEvan, line 262: Should this read `subprocess.CalledProcessError`?

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+    verbose=False,
+):
+    """
+    Performs diamond blastp search using query sequence against diamond formatted database
 
     Parameters
     ----------
     fasta : str
-        </path/to/fasta/file>. May be amino acid or nucleotide sequences
+        Path to fasta file having the query sequence. May be amino acid or nucleotide sequences
```
WiscEvan, line 295: The _fasta_ may contain multiple sequences. These would all be query sequences. The description suggests the fasta file should only contain one query sequence, which is false.

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+        '</path/to/fasta/file>' 
     database : str
-        </path/to/diamond/formatted/database>.
+        Path to diamond formatted database
+        '</path/to/diamond/formatted/database>'
```
WiscEvan, line 299: I'm not sure having both of these lines is helpful.

```suggestion

```

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
-    cpus : int
-        number of cpus to use (the default is 1).
-    tmpdir : type
-        </path/to/temporary/directory> (the default is os.curdir).
-    force : boolean
-        overwrite existing diamond results `force` (the default is False).
-    verbose : boolean
-        log progress to terminal `verbose` (the default is False).
+        Path to output file
+        '</path/to/output/file>'
```
WiscEvan, line 302: I'm not sure having both of these lines is helpful.

```suggestion

```

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+    blast_type : str, optional
+        blastp to align protein query sequences against a protein reference database,
+        blastx to align translated DNA query sequences against a protein reference database, by default 'blastp'
+    evalue : float, optional
+        cutoff e-value to count hit as significant,, by default float('1e-5')
```
WiscEvan, line 307:
```suggestion
        cutoff e-value to count hit as significant, by default float('1e-5')
```

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+    maxtargetseqs : int, optional
+        max number of target sequences to retrieve per query by diamond, by default 200
+    nproc : int, optional
+        Number of processors to be used, by default uses all the processors of the system
+    tmpdir : str, optional
+        Path to temporary directory
+        '</path/to/temporary/directory>' by default os.curdir
```
WiscEvan, line 314: I'm not sure having both of these lines is helpful.

```suggestion
        Path to temporary directory, by default os.curdir
```

The default should probably _not_ be the current directory, but rather an actual temp directory. You could probably implement this using the [tempfile](https://docs.python.org/3/library/tempfile.html) module

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
+    verbose : bool, optional
+        log progress to terminal, by default False
 
     Returns
     -------
     str
-        `outfpath`
+        Path to BLAST results
 
     Raises
```
WiscEvan, line 325: You need to update the docstring to reflect that a `subprocess.CalledProcessError` is raised.

autometa/common/external/diamond.py
```diff
@@ -182,50 +337,60 @@ def blast(fasta, database, outfpath, blast_type='blastp', evalue=float('1e-5'),
...
     Parameters
     ----------
     results : str
         </path/to/blastp/outfmt6/output/file>
-    top_pct : 0 < float <= 1
-        bitscore filter applied to each qseqid (the default is 0.9).
+    bitscr_thresh : 0 < float <= 1, optional
+        bitscore filter applied to each qseqid, by default 0.9
+        Used to determine whether the bitscore is above a threshold value. 
+        For example, if it is 0.9 then only bitscores >= 0.9 * the top bitscore are accepted
```
WiscEvan, line 391: This would be applied to the sseqids. This could probably be more concise/clear

autometa/common/external/diamond.py
```diff
@@ -283,18 +452,28 @@ def parse(results, top_pct=0.9, verbose=False):
...
-                hits.update({hit.qseqid:hit})
+                hits.update({hit.qseqid: hit})
                 topbitscore = bitscore
                 temp = set([hit.qseqid])
                 continue
-            if bitscore >= top_pct * topbitscore:
+            if bitscore >= bitscr_thresh * topbitscore:
+                import pdb
+
+                pdb.set_trace()
```
WiscEvan, line 465: `pdb` should not be in the code
```suggestion
```

autometa/common/external/diamond.py
```diff
@@ -343,78 +518,104 @@ def add_taxids(hits, database, verbose=True):
...
+    for qseqid, hit in tqdm(
+        hits.items(), disable=disable, total=n_qseqids, desc=desc, leave=False
+    ):
         for sseqid in hit.sseqids:
             taxid = acc2taxids.get(sseqid)
-            # if taxid is None:
-            #     raise DatabasesOutOfDateError(f'{sseqid} deprecated/suppressed/removed')
-            hit.sseqids[sseqid].update({'taxid':taxid})
+            if not taxid:
+                taxid = ncbi.merged.get(taxid, taxid)
```
WiscEvan, line 551: If the taxid is `None`, how are you searching for this in merged? This appears to be a bug

autometa/common/external/diamond.py
```diff
@@ -343,78 +518,104 @@ def add_taxids(hits, database, verbose=True):
...
+            acc2taxids.update({acc_num: taxid})
         if acc_ver in accessions:
-            acc2taxids.update({acc_ver:taxid})
+            acc2taxids.update({acc_ver: taxid})
     fh.close()
     n_qseqids = len(hits)
-    desc = f'Translating {n_qseqids} qseqids\' accessions to taxids'
-    for qseqid,hit in tqdm(hits.items(), disable=disable, total=n_qseqids, desc=desc, leave=False):
+    desc = f"Translating {n_qseqids} qseqids' accessions to taxids"
+    ncbi = NCBI(NCBI_DIR, verbose)  # Instantiating the NCBI class
```
WiscEvan, line 544: Is this necessary at this stage?

autometa/common/external/diamond.py
```diff
@@ -343,78 +518,104 @@ def add_taxids(hits, database, verbose=True):
...
+    )
+    parser.add_argument(
+        "--nproc", help="number of processors to use", default=mp.cpu_count(), type=int
+    )
+    parser.add_argument("--tmpdir", help="</path/to/tmp/directory>", default=os.curdir)
+    parser.add_argument(
+        "--bitscr-thresh",
+        help="threshold above which bitscores will be selected",
+        default=0.9,
+    )
```
WiscEvan, line 592: This is an end-user argument and should be more descriptive. e.g. `--bitscore-threshold`. The help text is not very informative

autometa/common/external/diamond.py
```diff
@@ -343,78 +518,104 @@ def add_taxids(hits, database, verbose=True):
...
+        "--evalue", help="diamond evalue threshold", default=float("1e-5")
+    )
+    parser.add_argument(
+        "--maxtargetseqs",
+        help="max target sequences to retrieve per query",
+        default=200,
+        type=int,
+    )
+    parser.add_argument(
+        "--nproc", help="number of processors to use", default=mp.cpu_count(), type=int
```
WiscEvan, line 585: Do you think `nproc` or `cpus` is more clear to the end user?

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
 
-from autometa.common.utilities import make_pickle,file_length
+from autometa.common.utilities import make_pickle, file_length
 from autometa.common.external import prodigal
+from autometa.taxonomy.ncbi import NCBI
 
 logger = logging.getLogger(__name__)
 
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+NCBI_DIR = os.path.join(BASE_DIR, "databases", "ncbi")
```
WiscEvan, line 44: I'm not sure this is necessary.

autometa/common/external/diamond.py
```diff
@@ -29,146 +29,301 @@
...
 
+import multiprocessing as mp
+
 from itertools import chain
 from tqdm import tqdm
 
-from autometa.common.utilities import make_pickle,file_length
+from autometa.common.utilities import make_pickle, file_length
 from autometa.common.external import prodigal
+from autometa.taxonomy.ncbi import NCBI
```
WiscEvan, line 39: I do not think this is necessary

autometa/common/external/diamond.py
```diff
@@ -182,50 +337,60 @@ def blast(fasta, database, outfpath, blast_type='blastp', evalue=float('1e-5'),
...
+    try:
+        proc.check_returncode()
+    except subprocess.CalledProcessError as err:
+        raise err
     return outfpath
 
-def parse(results, top_pct=0.9, verbose=False):
-    """Retrieve diamond results from output table
+
+def parse(results, bitscr_thresh=0.9, verbose=False):
```
WiscEvan, line 380: I'm not sure `bitscr_thresh` is clear as a variable name. When inspecting blast results, the bitscore is usually a float that can range to the 1000s. However, we are asking for a float between 0 and 1. I think this is confusing.

autometa/common/external/diamond.py
```diff
@@ -283,18 +452,28 @@ def parse(results, top_pct=0.9, verbose=False):
...
+
+                pdb.set_trace()
                 hits[hit.qseqid] += hit
     return hits
 
+
 def add_taxids(hits, database, verbose=True):
-    """Translates accessions to taxid translations from prot.accession2taxid.gz
+    """
+    Translates accessions to taxid translations from prot.accession2taxid.gz
```
WiscEvan, line 472: This translates accessions to taxids. Not taxid translations

autometa/common/external/diamond.py
```diff
@@ -283,18 +452,28 @@ def parse(results, top_pct=0.9, verbose=False):
...
+    If an accession number is no longer available (either due to being suppressed, 
+    deprecated or removed by NCBI), then merged.dmp is searched. If not found even there
+    then None is returned.
```
WiscEvan, line 475: If you can not retrieve a taxid, how are you able to search merged?

autometa/common/external/diamond.py
Sidduppal: I understand that this is a biit redundant but I was thinking we should keep this to maintain consistency. Let me know what you think

autometa/common/external/diamond.py
Sidduppal: I understand that this is a biit redundant but I was thinking we should keep this to maintain consistency. Let me know what you think

autometa/common/external/diamond.py
Sidduppal: `cpus` would be better. Do you thin we should change this in other scripts as well?

autometa/common/external/diamond.py
Sidduppal: I added these to used `self.merged` here:
https://github.com/Sidduppal/Autometa/blob/72fc07b3dd0d3503f71da7447b04aa653b0c85e6/autometa/common/external/diamond.py#L551-L552

But after this [comment](https://github.com/KwanLab/Autometa/pull/87#discussion_r436074791), it looks like my solution was not correct after all. I will re-think how to search the `merge.dmp`. any suggestions?

autometa/common/external/diamond.py
Sidduppal: From my understanding of the script and trying with `pdb` I found that `hit.sseqids` will always have a single `sseqid`, and thus we only need to remove the first key. Please let me know in case I am missing something.

autometa/common/external/diamond.py
```diff
@@ -582,12 +555,14 @@ def main():
...
+    )
+    parser.add_argument(
+        "--tmpdir", help="</path/to/tmp/directory>", default=tempfile.mkdtemp()
     )
-    parser.add_argument("--tmpdir", help="</path/to/tmp/directory>", default=os.curdir)
     parser.add_argument(
-        "--bitscr-thresh",
-        help="threshold above which bitscores will be selected",
+        "--top_percentile",
+        help="pecentile above which hits will be retreived",
```
WiscEvan, line 565:
```suggestion
        "--top-percentile",
        help="percentile above which hits will be retrieved",
```

autometa/common/external/diamond.py
```diff
@@ -582,12 +555,14 @@ def main():
         type=int,
     )
     parser.add_argument(
-        "--nproc", help="number of processors to use", default=mp.cpu_count(), type=int
+        "--cpus", help="number of processors to use", default=mp.cpu_count(), type=int
+    )
+    parser.add_argument(
+        "--tmpdir", help="</path/to/tmp/directory>", default=tempfile.mkdtemp()
```
WiscEvan, line 561: Depending on how you handle the comments about the `tmpdir` above, you may be able to remove this.

autometa/common/external/diamond.py
WiscEvan: This may be true for when parsing the blast results, but this may not be true in other circumstances. Only removing one when instead the behavior intends to remove all, could result in a difficiult to find "_gotcha_" kind of 🐛

autometa/common/external/diamond.py
WiscEvan: I think probably only necessary when it is front facing. But kind of a judgement call. Which do you prefer? Either works for me.

WiscEvan:
It appears the docs are failing from the `tempdir` argument. The default value you are setting as `tempfile.mkdtemp` when `tempfile is not imported. This should be resolved if this parameter ends up being removed. See https://github.com/KwanLab/Autometa/pull/87#discussion_r439560687

autometa/common/external/diamond.py
Sidduppal: I think `cpus` makes it clear to the end-user as well as the developer. 
We can definitely have a different front-facing variable but it may cause some minor confusions in the future.

autometa/common/external/diamond.py
Sidduppal: Why not use underscores (`_`)?

autometa/common/external/diamond.py
```diff
@@ -330,6 +307,8 @@ def blast(
         provided `blast_type` is not 'blastp' or 'blastx'
     OSError
         Diamond execution failed
+    subprocess.CalledProcessError
+        Failed to run blast
     """
     if not os.path.exists(fasta):
         raise FileNotFoundError(fasta)
```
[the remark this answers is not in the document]
Sidduppal: 1. I think `if verbose` can be a handy tool if someone is running the script as a standalone module. What do you think about it? 
Else I can remove the `if verbose` statements and just use the logging statement for all the functions.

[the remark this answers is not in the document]
Sidduppal: I think we should keep `tmpdir` as an input from the user. It may be the case that the user needs to specify a different directory rather than the temporary directory like in the case of space issues. I don't think we should use a `tempdir` flag with 1diamond` if we are randomly creating a temporary directory, as `diamond` itself creates a temporary directory to use. 
From what I read from the diamond documentation the `tempdir` flag is only needed in cases where the output directory does not has enough space available.
Let me know what you think about it

Sidduppal:
I was reading about the `assert` method. It will raise an error in case the query sequences are not equal. Do we want an error to be raised? If yes, should be mention that in the docstrings?

WiscEvan:
> I was reading about the `assert` method. It will raise an error in case the query sequences are not equal. Do we want an error to be raised? If yes, should be mention that in the docstrings?

You make a good point. What do you think? I think arguments could be made for either implementation.

autometa/common/external/diamond.py
```diff
@@ -28,204 +28,347 @@
...
-    if proc.returncode:
-        raise OSError(f'DiamondFailed:\nArgs:{proc.args}\nReturnCode:{proc.returncode}')
+    proc = subprocess.run(
+        cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, shell=False
+    )
+    try:
+        proc.check_returncode()
+    except subprocess.CalledProcessError as err:
+        raise err
+    shutil.rmtree(tmpdir, ignore_errors=True)
```
WiscEvan, line 354: Have you tested this? What happens if the user were to specify the current working directory as their tmpdir? What about the autometa output directory? Do we want to remove everything in this directory?

docs/parse_argparse.py
```diff
@@ -57,7 +57,7 @@ def get_argparse_block(fpath):
         argparse block between (and includes) import argparse and args = parser.parse_args() to be written to a temporary file
     """
     writing = False
-    outlines = "import os\nimport multiprocessing as mp\n"
+    outlines = "import os\nimport multiprocessing as mp\nimport tempfile\n"
```
WiscEvan, line 60: This seems kind of like a heavy handed approach for one specific file. I'm not sure this needs to be included in every script.

docs/parse_argparse.py
```diff
@@ -105,30 +103,38 @@ def get_usage(argparse_lines):
     -------
     wrapped_lines : string
         indented arparse output after running the `--help` command
+
+    Raises
+    ------
+    CalledProcessError
```
WiscEvan, line 109:
```suggestion
    subprocess.CalledProcessError
```

autometa/common/external/diamond.py
```diff
@@ -343,35 +492,82 @@ def add_taxids(hits, database, verbose=True):
...
+        "--maxtargetseqs",
+        help="max target sequences to retrieve per query",
+        default=200,
+        type=int,
+    )
+    parser.add_argument(
+        "--cpus", help="number of processors to use", default=mp.cpu_count(), type=int
+    )
+    parser.add_argument(
+        "--tmpdir", help="</path/to/tmp/directory>", default=tempfile.mkdtemp()
```
WiscEvan, line 559: Is placing a default value necessary here? The function (`blastp`) already places a default value for `tmpdir`. This seems like it is adding more headache and not really doing anything for the end user.

Sidduppal:
> 

> > I was reading about the `assert` method. It will raise an error in case the query sequences are not equal. Do we want an error to be raised? If yes, should be mention that in the docstrings?
> 
> You make a good point. What do you think? I think arguments could be made for either implementation.

We should keep `assert` and raise an error if the condition is False. However, that should also be mentioned in the docstring. From my understanding, the error would not be raised if the script is being used as the part of the pipeline as the following lines (that are above `hits[hit.qseqid] += hit`) make sure that the query sequences are the same ones.

https://github.com/KwanLab/Autometa/blob/ddc2bf951d6c4a8c29e9e2301e15a762327b0267/autometa/common/external/diamond.py#L287-L291

Raising an error would be good when running the script as a stand-alone module or for future developers to test the module.
Let me know what you think about this.

autometa/common/external/diamond.py
```diff
@@ -28,204 +28,359 @@
...
+            Query sequences are not equal
+        """
+        assert (
+            self == other_hit
+        ), f"qseqids do not match! {self.qseqid} & {other_hit.qseqid}"
+        for key in list(other_hit.sseqids.keys()):
+            try:
+                self.sseqids.pop(key)
+            except KeyError as err:
+                raise err
```
WiscEvan, line 214: A few problems here.

1. There is no need to use the `list` function when also employing the `dict.keys` method. When iterating through a dictionary, the default behavior is to iterate through the keys. So the method and the function are unnecessary.
2. It would be more helpful to be descriptive with your variables.  i.e. `sseqid` rather than `key`.
2. The error message is not helpful and this is 3 extra lines to give the same exact message as just brute forcing `self.sseqids.pop(key)`
3. I think the `try` and `except` block could be useful here, but you need to provide more helpful information.

```suggestion
        for sseqid in other_hit.sseqids:
            try:
                self.sseqids.pop(sseqid)
            except KeyError:
                raise KeyError(<'Some useful error message here'>)
```

autometa/common/external/diamond.py
```diff
@@ -28,204 +28,359 @@
...
+    """
+    cmd = f"diamond makedb --in {fasta} --db {database} -p {cpus}"
+    logger.debug(f"{cmd}")
+    proc = subprocess.run(
+        cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, shell=True
+    )
+    try:
+        proc.check_returncode()
+    except subprocess.CalledProcessError as err:
+        raise err
```
WiscEvan, line 263: 1. Since you returning the same error message that is being returned from `subprocess.CalledProcessError`, you could place this within `subprocess.run(check=True)`
2. You are running `subprocess.run` with `shell=True`, this has been [documented as being a security risk](https://docs.python.org/3/library/subprocess.html#security-considerations). So we should try to minimize this as much as possible.
```suggestion
    cmd = ["diamond", "makedb", "--in", fasta, "--db", database ,"-p", cpus]
    subprocess.run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
    logger.debug(" ".join(cmd))
```

autometa/common/external/diamond.py
```diff
@@ -28,204 +28,359 @@
...
-        proc = subprocess.run(cmd, stdout=stdout, stderr=stderr)
-    if proc.returncode:
-        raise OSError(f'DiamondFailed:\nArgs:{proc.args}\nReturnCode:{proc.returncode}')
+    proc = subprocess.run(
+        cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, shell=False
+    )
+    try:
+        proc.check_returncode()
+    except subprocess.CalledProcessError as err:
+        raise err
```
WiscEvan, line 366: 1. Similar comments as in the `makedatabase` function.
2. `shell=False` is the default and not necessary

```suggestion
    subprocess.run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
```

autometa/common/external/diamond.py
```diff
@@ -343,35 +504,80 @@ def add_taxids(hits, database, verbose=True):
...
+
+def main():
+    import argparse
+    import os
+
+    parser = argparse.ArgumentParser(
+        description="""
+    Retrieves blastp hits with provided input assembly
+    """
+    )
```
WiscEvan, line 547: Since you are specifying multiple defaults in the argparse parser. You should apply the defaults formatter.

https://github.com/KwanLab/Autometa/blob/ddc2bf951d6c4a8c29e9e2301e15a762327b0267/docs/template.py#L89-L94

```suggestion
    parser = argparse.ArgumentParser(
        description="Retrieves blastp hits with provided input assembly",
        formatter_class=argparse.ArgumentDefaultsHelpFormatter
    )
```

Sidduppal:
I think I made a mistake in the last commit by removing the default value of `tempdir`. Doing this results in `tmpdir=args.tmpdir` being equal to `None` which is an invalid input for diamond blast. From how I look at it we have three solutions:
1. Set the default value as `os.curdir`, so we don't have to add anything more in `parse.argparse.py`
2. Hardcode `import tempfile` in `parse.argaprse.py`
3. Write `import tempfile` just below `import argparse` in `diamond.py`
I was more leaning towards the third option as it is more versatile, but I would like to hear your views on it as well.

WiscEvan:
> I think I made a mistake in the last commit by removing the default value of `tempdir`. Doing this results in `tmpdir=args.tmpdir` being equal to `None` which is an invalid input for diamond blast. From how I look at it we have three solutions:
> 
> 1. Set the default value as `os.curdir`, so we don't have to add anything more in `parse.argparse.py`
> 2. Hardcode `import tempfile` in `parse.argaprse.py`
> 3. Write `import tempfile` just below `import argparse` in `diamond.py`
>    I was more leaning towards the third option as it is more versatile, but I would like to hear your views on it as well.

Why not just not pass in `args.tmpdir` if it is not specified? This seems like the simplest solution

Sidduppal:
> > I think I made a mistake in the last commit by removing the default value of `tempdir`. Doing this results in `tmpdir=args.tmpdir` being equal to `None` which is an invalid input for diamond blast. From how I look at it we have three solutions:
> > 
> > 1. Set the default value as `os.curdir`, so we don't have to add anything more in `parse.argparse.py`
> > 2. Hardcode `import tempfile` in `parse.argaprse.py`
> > 3. Write `import tempfile` just below `import argparse` in `diamond.py`
> >    I was more leaning towards the third option as it is more versatile, but I would like to hear your views on it as well.
> 
> Why not just _not_ pass in `args.tmpdir` if it is not specified? This seems like the simplest solution

Do you mean something like:
```python
tempdir=tempfile.mkdtemp() if args.tempdir==None else args.tempdir 
```

WiscEvan:
You need to update this PR with the most recent changes from `dev`

Sidduppal:
Yes, but I didn't push/commit anything. I was thinking of resolving the issue of tempdir and then pushing. Any ideas on which option to go with?

WiscEvan:
> > > I think I made a mistake in the last commit by removing the default value of `tempdir`. Doing this results in `tmpdir=args.tmpdir` being equal to `None` which is an invalid input for diamond blast. From how I look at it we have three solutions:
> > > 
> > > 1. Set the default value as `os.curdir`, so we don't have to add anything more in `parse.argparse.py`
> > > 2. Hardcode `import tempfile` in `parse.argaprse.py`
> > > 3. Write `import tempfile` just below `import argparse` in `diamond.py`
> > >    I was more leaning towards the third option as it is more versatile, but I would like to hear your views on it as well.
> > 
> > 
> > Why not just _not_ pass in `args.tmpdir` if it is not specified? This seems like the simplest solution
> 
> Do you mean something like:
> `tmpdir=not args.tmpdir if args.tmpdir == None else args.tmpdir`
> This is also returning error. It needs a directory as input and not a boolean

`tmpdir=not args.tmpdir if args.tmpdir == None else args.tmpdir` is too verbose and will result in a boolean for one of the values.

More along the lines of what you are passing in to your function.

I was thinking of when you pass `args.tmpdir` into the function:

1. 
```python

tmpdir = tempfile.mkdtemp() if not args.tmpdir else args.tmpdir
result = blast(
        fasta=args.fasta,
        database=args.database,
        outfpath=args.outfile,
        blast_type=args.blast_type,
        evalue=args.evalue,
        maxtargetseqs=args.maxtargetseqs,
        cpus=args.cpus,
        tmpdir=tmpdir,
        force=args.force,
        verbose=args.verbose)
```

2. 
```python
if not args.tmpdir:
    result = blast(
        fasta=args.fasta,
        database=args.database,
        outfpath=args.outfile,
        blast_type=args.blast_type,
        evalue=args.evalue,
        maxtargetseqs=args.maxtargetseqs,
        cpus=args.cpus,
        force=args.force,
        verbose=args.verbose)
else:
    result = blast(
        fasta=args.fasta,
        database=args.database,
        outfpath=args.outfile,
        blast_type=args.blast_type,
        evalue=args.evalue,
        maxtargetseqs=args.maxtargetseqs,
        cpus=args.cpus,
        tmpdir=args.tmpdir,
        force=args.force,
        verbose=args.verbose)
```

Both of these routines seem round about though. Is having `tempfile.mkdtemp()` as a default necessary?

From the [diamond manual](http://gensoft.pasteur.fr/docs/diamond/0.8.29/diamond_manual.pdf):

>3.6 Memory & performance options
>
>--tmpdir/-t <directory>
Directory to be used for temporary storage. This is set to the output directory by default.
The amount of disk space that will be used depends on the program’s settings and your
data. As a general rule you should ensure that 100 GB of disk space are available here.
If you run the program in a cluster environment, and disk space is only available over a
slow network based file system, you may want to set the --tmpdir option to /dev/shm. This
will keep temporary information in memory and thus increase the program’s memory usage
substantially.

autometa/common/external/diamond.py
```diff
@@ -28,204 +28,359 @@
...
-        '--out',
+        "--outfmt",
+        "6",
+        "--out",
         outfpath,
-        '--tmpdir',
-        tmpdir]
+        "--tmpdir",
+        tmpdir,
+    ]
```
WiscEvan, line 355: It seems to me, the only time you will need to add `--tmpdir tmpdir` to the command is when this is supplied by the user. Otherwise, I think diamond will handle this. Similarly, you could place the logic _within_ this function if you would like to set the tmpdir using the `tempfile.mkdtemp` method. The default `tmpdir` parameter could be `None`. If the default were to be `None`, then you could assign the `tmpdir` variable to `tempfile.mkdtemp()`.

autometa/common/external/diamond.py
```diff
@@ -157,66 +269,72 @@ def blast(
...
     """
     if not os.path.exists(fasta):
         raise FileNotFoundError(fasta)
     if os.path.exists(outfpath) and not force:
-        empty = not os.stat(outfpath).st_size
-        if not empty:
-            if verbose:
-                logger.warning(f"FileExistsError: {outfpath}. To overwrite use --force")
+        if os.path.getsize(outfpath) and verbose:
+            logger.warning(f"FileExistsError: {outfpath}. To overwrite use --force")
```
WiscEvan, line 326: Why is returning here dependent on the user specifying verbose?

```suggestion
        if os.path.getsize(outfpath):
            logger.warning(f"FileExistsError: {outfpath}. To overwrite use --force")
```

autometa/common/external/diamond.py
```diff
@@ -157,66 +269,72 @@ def blast(
...
             return outfpath
     blast_type = blast_type.lower()
     if blast_type not in ["blastp", "blastx"]:
         raise ValueError(f"blast_type must be blastp or blastx. Given: {blast_type}")
+    if tmpdir == None:
+        tmpdir = tempfile.mkdtemp()
     if verbose:
-        logger.debug(f"Diamond{blast_type.title()} {fasta} against {database}")
+        logger.debug(f"Diamond{blast_type} {fasta} against {database}")
```
WiscEvan, line 334: minor changes for whitespace and executable

```suggestion
        logger.debug(f"diamond {blast_type} {fasta} against {database}")
```

autometa/common/external/diamond.py
```diff
@@ -405,43 +527,19 @@ def add_taxids(hits, database, verbose=True):
...
-    hits_fname = ".".join([fname, "pkl.gz"])
-    hits_fpath = os.path.join(dirpath, hits_fname)
-    pickled_fpath = make_pickle(obj=hits, outfpath=hits_fpath)
-    logger.debug(f"{len(hits):,} diamond hits serialized to {pickled_fpath}")
-
-
-if __name__ == "__main__":
-    import argparse
+def main():
     import os
```
WiscEvan, line 535: This appears to be imported twice?

```suggestion

```

autometa/common/external/diamond.py
```diff
@@ -157,66 +269,72 @@ def blast(
...
-            if verbose:
-                logger.warning(f"FileExistsError: {outfpath}. To overwrite use --force")
+        if os.path.getsize(outfpath) and verbose:
+            logger.warning(f"FileExistsError: {outfpath}. To overwrite use --force")
             return outfpath
     blast_type = blast_type.lower()
     if blast_type not in ["blastp", "blastx"]:
         raise ValueError(f"blast_type must be blastp or blastx. Given: {blast_type}")
+    if tmpdir == None:
+        tmpdir = tempfile.mkdtemp()
```
WiscEvan, line 332: Typically if you are checking if `something` is `None` you can just write `if something`.
You could simplify this to: `tmpdir = tmpdir if tmpdir else tempfile.mkdtemp()`.

That being said, this overwrites the option for diamond to run its default tmpdir. I think if `tmpdir` is `None` then the command should not be added from the list of commands or could be removed from the list of commands.

```python
    cmd = [...]
    if tmpdir:
        cmd.extend(["--tmpdir", tmpdir])
```

Sidduppal:
I have addressed the comment regarding using the default temporary directory (specified by diamond) when the user does not specify anything. However, I was thinking this will only work when the script is being run as a standalone module. When the script will be run as the part of the pipe, the parameter, `tmpdir=tempfile.mkdtemp()` in the function definition would create a temporary directory. I don't think this is a major issue but I just wanted to bring your attention to it.

autometa/common/external/diamond.py
```diff
@@ -462,14 +558,43 @@ def main(args):
         default=200,
         type=int,
     )
-    parser.add_argument("--cpus", help="num cpus to use", default=0, type=int)
-    parser.add_argument("--tmpdir", help="</path/to/tmp/directory>", default=os.curdir)
     parser.add_argument(
-        "--top-pct", help="top percentage of hits to retrieve", default=0.9
+        "--cpus", help="number of processors to use", default=mp.cpu_count(), type=int
+    )
+    parser.add_argument("--tmpdir", help="</path/to/tmp/directory>")
```
WiscEvan, line 564: The help text for this could be more descriptive. It is unclear why specifying a tmp directory is necessary. What is using the tmp directory?

Something like "Path to tmp directory used by diamond (See diamond manual section 3.6 for details)."

autometa/common/external/diamond.py
```diff
@@ -157,66 +269,71 @@ def blast(
     blast_type="blastp",
     evalue=float("1e-5"),
     maxtargetseqs=200,
-    cpus=0,
-    tmpdir=os.curdir,
+    cpus=mp.cpu_count(),
+    tmpdir=tempfile.mkdtemp(),
     force=False,
     verbose=False,
 ):
```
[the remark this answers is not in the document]
Sidduppal: You are right, the default doesn't have to be `tempfile.mkdtemp()`. Looks like I made a silly mistake by not altering it already :facepalm: The default is now `None` thanks for correcting it :+1: :clinking_glasses:

autometa/common/external/diamond.py
```diff
@@ -108,29 +116,110 @@ def __init__(
...
+        assert (
+            self == other_hit
+        ), f"qseqids do not match! {self.qseqid} & {other_hit.qseqid}"
+        for sseqid in other_hit.sseqids.keys():
+            try:
+                self.sseqids.pop(sseqid)
+            except KeyError:
+                raise KeyError(
+                    "The query sequence corresponsing to the DiamondResult is not present!"
+                )
```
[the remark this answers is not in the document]
Sidduppal: Thanks for the link I was not aware of that :beers: :bowing_man: :checkered_flag:

autometa/common/external/diamond.py
```diff
@@ -231,28 +347,33 @@ def blast(
...
-        raise OSError(f"DiamondFailed:\nArgs:{proc.args}\nReturnCode:{proc.returncode}")
+    subprocess.run(
+        cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True
+    )
     return outfpath
 
 
-def parse(results, top_pct=0.9, verbose=False):
-    """Retrieve diamond results from output table
+def parse(results, top_percentile=0.9, verbose=False):
```
jason-c-kwan, line 363: You ignored my request in issue #36. Please read it again:

>In the script, there is an argument/variable called top_pct or a variant of this name. This is a float value between 0 and 1 (although it can't exactly be zero), used to determine if the bitscore of hits is above a threshold. For example, if it is 0.9 then only bitscores >= 0.9 * the top bitscore are accepted. The problem here is not one of programmatic logic, but rather readability for contributors. top_pct looks like "top percentage", which would suggest, for example, that for a value of 0.9, 90% will be kept. However, the opposite is true in this case - only 10% will be kept. This is compounded by the help message in the if __name__ == '__main__': block, which describes this argument as "top percentage of hits to retrieve". So what I am saying is that the variable name and documentation is confusing. Perhaps it could be called "bitscore_tolerance" or something? Accompanied by a change in documentation and inclusion of explanatory comments.

The variable does not describe a "top percentile" to keep, but rather some sort of threshold.

autometa/common/external/diamond.py
```diff
@@ -347,7 +472,10 @@ def add_taxids(hits, database, verbose=True):
     hits : dict
         {qseqid: DiamondResult, ...}
     database : str
-        </path/to/prot.accession2taxid.gz>
+        Path to prot.accession2taxid.gz database
+        e.g. </path/to/prot.accession2taxid.gz>
```
jason-c-kwan, line 476: This line is not needed.

autometa/common/external/diamond.py
```diff
@@ -462,14 +557,46 @@ def main(args):
...
+        "--cpus", help="number of processors to use", default=mp.cpu_count(), type=int
+    )
+    parser.add_argument(
+        "--tmpdir",
+        help="Path to directory which will be used for temporary storage by diamond",
+    )
+    parser.add_argument(
+        "--top-percentile",
+        help="percentile above which hits will be retrieved",
+        default=0.9,
```
jason-c-kwan, line 570: See other comment

autometa/common/external/diamond.py
Sidduppal: I am not sure which line(s) are you pointing towards.
Also, I kept both the verbal description as well as the example to maintain consistency throughout the codebase.
Let me know if you only need the description, I would the remove the "e.g." from other function like `blast` as well

jason-c-kwan: Line 476. It doesn't add any information that is not in line 475, especially because you specify the exact filename in line 475.

autometa/common/external/diamond.py
Sidduppal: I was thinking of`bitscr_filter` with the corresponding `help` text:
"hits with bitscore greater than bitscr-filter*top bitscore will be kept".
Other options that I thought were `bitscore_threshold` and the one you suggested i.e. `bitscr_tolerance`. I was moe inclined towards `bitscr_filter`.
What do you think about it? Any ideas?

jason-c-kwan: `bitscore_filter` is OK, but can you not abbreviate? If you have a decent text editor it should autocomplete anyway.

Sidduppal: Alright!
sure I can write the full name. I don't know why I was of the view that developers prefer variable names that have less character (i.e. abbreviated), seems like I was wrong. I'll make the change and push the changes!

autometa/common/external/diamond.py
```diff
@@ -42,12 +42,18 @@
 
 class DiamondResult:
     """
-    DiamondResult class for handling of Diamond result's table
+    DiamondResult class to instantiate a DiamondResult object for each qseqid.
+    Allows for better handling of subject sequence IDs that hit to a qseqid.
+
+    Includes methods used to modify the DiamondResult object (add or remove sseqid from sseqids dictionary),
+    check if two DiamondResult object have the same qseqid and return a user firendly and unambiguous output
```
jason-c-kwan, line 49: Grammatical and spelling error:

```suggestion
    check if two DiamondResult objects have the same qseqid and return a user friendly and unambiguous output
```

autometa/common/external/diamond.py
```diff
@@ -42,12 +42,18 @@
...
+    from str() ad repr() respectively. Also used to returns the sseqid with the highest bitscore amongst all
+    the subject sequences that hit a query. These methds come in handy when retrieving diamond results from output table.
```
jason-c-kwan, line 51: More spelling errors:

```suggestion
    from str() ad repr() respectively. Also used to return the sseqid with the highest bitscore amongst all
    the subject sequences that hit a query. These methods come in handy when retrieving diamond results from output table.
```

autometa/common/external/diamond.py
```diff
@@ -297,8 +301,7 @@ def blast(
     cpus : int, optional
         Number of processors to be used, by default uses all the processors of the system
     tmpdir : str, optional
-        Path to temporary directory
-        '</path/to/temporary/directory>' by default output directory
+        Path to temporary directory by default, output directory
```
jason-c-kwan, line 304: Should be two sentences.

```suggestion
        Path to temporary directory. By default, same as the output directory
```

Sidduppal: I was using VS code extension for docstring with the type set to `numpy`. Going through `numpy` docstring documentation I was not able to find a standard way to mention the default value, but since we all are now using VS code with the same extension, I think having "by default" in the same sentence would help maintain consistency. What do you think about this?

jason-c-kwan: No. I put this suggestion because the sentence structure you put is wrong. "Path to temporary directory by default" means that by default it is a temporary directory, and I suppose if the user chooses it will be some other type of directory? It doesn't make much sense. Then there is the other part of the sentence. "Path to temporary directory by default, output directory". The "output directory" part is out there on its own as if it was the start of the next clause, but then there is no verb or anything else. The punctuation I suggested makes it unambiguous that "By default" belongs with "output directory". I think your priority should be clarity before consistency. (But also I highly doubt that the numpy docstrings are grammatically wrong).

autometa/common/external/diamond.py
```diff
@@ -46,9 +46,9 @@ class DiamondResult:
     Allows for better handling of subject sequence IDs that hit to a qseqid.
 
     Includes methods used to modify the DiamondResult object (add or remove sseqid from sseqids dictionary),
-    check if two DiamondResult object have the same qseqid and return a user firendly and unambiguous output
-    from str() ad repr() respectively. Also used to returns the sseqid with the highest bitscore amongst all
-    the subject sequences that hit a query. These methds come in handy when retrieving diamond results from output table.
+    check if two DiamondResult objects have the same qseqid and return a user friendly and unambiguous output
+    from str() ad repr() respectively. Also used to return the sseqid with the highest bitscore amongst all
```
jason-c-kwan, line 50: Found another typo:
```suggestion
    from str() and repr() respectively. Also used to return the sseqid with the highest bitscore amongst all
```

autometa/common/external/diamond.py
jason-c-kwan: Please address this before we merge the PR.

PR #96, opened 2020-06-21
Rank-specific binning

WiscEvan:
:art: Add taxonomy specific splitting control
:art::racehorse: Add reverse-ranks parameter
:art::racehorse: Add starting-rank parameter

### DIFF
--- a/autometa/binning/recursive_dbscan.py
+++ b/autometa/binning/recursive_dbscan.py
@@ -704,7 +704,7 @@ def main():
         action="store_true",
         default=False,
         help="Reverse order at which to split taxonomy by canonical-rank."
-        " Contigs will be split in order of species, genus, family, order, class, phylum, superkingdom.",
+        " When --reverse-ranks given, contigs will be split in order of species, genus, family, order, class, phylum, superkingdom.",
     )
     parser.add_argument(
         "--domain",
--- a/autometa/binning/recursive_dbscan.py
+++ b/autometa/binning/recursive_dbscan.py
@@ -703,8 +703,8 @@ def main():
         "--reverse-ranks",
         action="store_true",
         default=False,
-        help="Reverse canonical-rank taxonomy splitting"
-        " from superkingdom -> species to species -> superkingdom",
+        help="Reverse order at which to split taxonomy by canonical-rank."
+        " Contigs will be split in order of species, genus, family, order, class, phylum, superkingdom.",
     )
     parser.add_argument(
         "--domain",
--- a/autometa/binning/recursive_dbscan.py
+++ b/autometa/binning/recursive_dbscan.py
@@ -505,8 +505,9 @@ def binning(
     completeness=20.0,
     purity=90.0,
     taxonomy=True,
+    starting_rank="superkingdom",
     method="dbscan",
-    reverse=True,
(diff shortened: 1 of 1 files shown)


autometa/binning/recursive_dbscan.py
```diff
@@ -522,23 +523,26 @@ def binning(
...
+    method : str, optional
         Clustering `method` (the default is 'dbscan').
         choices = ['dbscan','hdbscan']
-    reverse : bool
-        True - [superkingdom,phylum,class,order,family,genus,species]
-        False - [species,genus,family,order,class,phylum,superkingdom]
-    verbose : bool
+    reverse_ranks : bool, optional
+        False - [superkingdom,phylum,class,order,family,genus,species] (Default)
+        True - [species,genus,family,order,class,phylum,superkingdom]
```
jason-c-kwan, line 544: Just checking - did you intent to make `reverse_ranks` have opposite meanings for `True` and `False` compared to `reverse`?

jason-c-kwan: It does seem to be internally consistent, but it is a little weird because later `NCBI.CANONICAL_RANKS` is reversed if `reverse_ranks == False`. This will probably trip contributors up.

autometa/binning/recursive_dbscan.py
```diff
@@ -678,6 +685,27 @@ def main():
...
+            "genus",
+            "species",
+        ],
+    )
+    parser.add_argument(
+        "--reverse-ranks",
+        action="store_true",
+        default=False,
+        help="Reverse canonical-rank taxonomy splitting"
+        " from superkingdom -> species to species -> superkingdom",
```
jason-c-kwan, line 707: I think this help message should more explicitly say what the order will be if `--reverse-ranks` is set to `True` vs. `False`, because as written, this will be confusing to users.

autometa/binning/recursive_dbscan.py
WiscEvan: Yes, I reversed the logic thinking this would be more intuitive to the end user. `NCBI.CANONICAL_RANKS` is by default ordered from most to least specific but as described in the NAR paper, the order splits by taxonomic rank from superkingdom down to species. The other way around this is to change the `NCBI.CANONICAL_RANKS` order and reverse the order in other files where it is used. I've placed comments at these lines of code to clarify to any developer.

autometa/binning/recursive_dbscan.py
```diff
@@ -703,8 +703,8 @@ def main():
         "--reverse-ranks",
         action="store_true",
         default=False,
-        help="Reverse canonical-rank taxonomy splitting"
-        " from superkingdom -> species to species -> superkingdom",
+        help="Reverse order at which to split taxonomy by canonical-rank."
+        " Contigs will be split in order of species, genus, family, order, class, phylum, superkingdom.",
     )
```
jason-c-kwan: This is still not completely clear. I think you will find users asking you exactly what order is --reverse-ranks and which is not.

```suggestion
        help="Reverse order at which to split taxonomy by canonical-rank."
        "When --reverse-ranks given, contigs will be split in order of species, genus, family, order, class, phylum, superkingdom.",
    )
```

PR #103, opened 2020-07-13
Samtools

Sidduppal:
:art: Simplified `samtools.py`. It now uses `subprocess.run` and does not require any unnecessary output or error files


autometa/common/external/samtools.py
```diff
@@ -42,49 +42,62 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+        ]
+        cmd_sort = [
+            "samtools",
+            "sort",
+            "-@",
+            str(cpus),
+            view_fpath,
+            "-o",
+            sort_fpath,
+        ]
```
WiscEvan, line 87: As far as I can tell, you are not piping the output of `samtools view ...` to `samtools sort ...`. This is not the same behavior as before.

Sidduppal: Yes, I specifically avoided piping as then we have to use `shell=True`, which posses a some security risks and is also discouraged.
https://docs.python.org/3/library/subprocess.html#security-considerations
https://stackoverflow.com/questions/13332268/how-to-use-subprocess-command-with-pipes

WiscEvan: You have linked to an example of how to pipe subprocess commands, but you did not implement this. Currently, you are not piping any of the output from view to sort. This should be performed as was previously implemented.

autometa/common/external/samtools.py
```diff
@@ -22,17 +22,16 @@
...
 import tempfile
 
-import pandas as pd
 import multiprocessing as mp
 
 logger = logging.getLogger(__name__)
 
 
-def sort(sam, bam, nproc=mp.cpu_count()):
+def sort(sam, bam, cpus=mp.cpu_count()):
```
WiscEvan, line 34: Have you tested this?

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,63 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-
-def main(args):
-    sort(args.sam, args.bam, args.nproc)
+        # See https://stackoverflow.com/questions/34147353/python-subprocess-chaining-commands-with-subprocess-run
+        # Popen starts the process and carries on while run starts it and waits for it to finish.
+        subprocess.run(
+            cmd_sort,
+            stdin=view.stdout,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
```
WiscEvan, line 93: Should both of these be piped? Should stderr be directed to `devnull`? Ideally if `samtools` fails, it would be useful to have this error information available to the user so they know _why_ samtools failed.

autometa/common/external/samtools.py
Sidduppal: I have tested this when running as a standalone module. I also changed `nproc` to `cpus` in the `coverage.py` script where sort was being used.

autometa/common/external/samtools.py
Sidduppal: You are right both should not be piped. 
I think we have 2 options to convey the error to the user:
- Create a `samtools.err` file. However, in all the other scripts we have just directed the `stderr` to the `DEVNULL`. 
- Log the error so that the user can look it up if he wants to.
I prefer the latter one as it maintains consistency and would reduce the creation of new files.

Sidduppal: Even though I have implemented to log the error after `sort` , I am not sure if `check=True` would allow that. I think `check=True` would raise an error and the logging would never happen. I could not verify this as I was unable to explicitly cause `sort` to crash.

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,65 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+        # Popen starts the process and carries on while run starts it and waits for it to finish.
+        sort = subprocess.run(
+            cmd_sort,
+            stdin=view.stdout,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+            check=True,
+        )
+        logger.error(sort.stderr)
+        shutil.move(sort_fpath, bam)
```
WiscEvan, line 98: Why not just remove `check=True` and handle the exception with `sort.check_returncode()`? Or wrap the `subprocess.run` in a `try/except` block?


Could also shorten the stackoverflow link as mentioned by Shaurya. There is a share button that creates a short link to the post.

```suggestion
        logger.error(view.stderr)
        # See https://stackoverflow.com/q/34147353/13118765
        # Popen starts the process and carries on while run starts it and waits for it to finish.
        try:
            sort = subprocess.run(
                cmd_sort,
                stdin=view.stdout,
                stdout=subprocess.DEVNULL,
                stderr=subprocess.PIPE,
                check=True,
            )
        except subprocess.CalledProcessError:
            logger.error(sort.stderr)
        shutil.move(sort_fpath, bam)
```


```suggestion
        logger.error(view.stderr)
        # See https://stackoverflow.com/q/34147353/13118765
        # Popen starts the process and carries on while run starts it and waits for it to finish.
        sort = subprocess.run(
            cmd_sort,
            stdin=view.stdout,
            stdout=subprocess.DEVNULL,
            stderr=subprocess.PIPE,
        )
        if sort.returncode:
            logger.error(sort.stderr)
            return None
        shutil.move(sort_fpath, bam)
```

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,65 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-def main(args):
-    sort(args.sam, args.bam, args.nproc)
+        logger.error(view.stderr)
+        # See https://stackoverflow.com/questions/34147353/python-subprocess-chaining-commands-with-subprocess-run
+        # Popen starts the process and carries on while run starts it and waits for it to finish.
+        sort = subprocess.run(
+            cmd_sort,
+            stdin=view.stdout,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
```
WiscEvan, line 94: Are you sure about redirecting standard output to `devnull`?

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,65 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
+        cmd_sort = [
+            "samtools",
+            "sort",
+            "-@",
+            str(cpus),
+            "-o",
+            sort_fpath,
+        ]
+        logger.debug(" ".join(cmd_view))
+        logger.debug(" ".join(cmd_sort))
```
WiscEvan, line 83: The `cmd_sort` command should not be logged until the process is being run. This could be confusing to the user as  `view = subprocess.Popen(cmd_view, stdout=subprocess.PIPE, stderr=subprocess.PIPE` may cause an error and `cmd_sort` has not yet been executed.

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,65 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-    samtools_out_dir = os.path.dirname(os.path.abspath(bam))
-    err = os.path.join(samtools_out_dir, "samtools.err")
-    out = os.path.join(samtools_out_dir, "samtools.out")
-    with tempfile.TemporaryDirectory() as tempdir:
-        temp_bam = os.path.join(tempdir, os.path.basename(bam))
-        cmd = (
-            f"samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {temp_bam}"
+    with tempfile.TemporaryDirectory() as tempdir:  # this will delete the temporary files even if program stops in between
+        sort_fpath = os.path.join(tempdir, os.path.basename(bam))
+        # See https://stackoverflow.com/questions/13332268/how-to-use-subprocess-command-with-pipes
```
WiscEvan, line 65: You can reduce the length of the url here for lookup later:

```suggestion
        # See https://stackoverflow.com/q/13332268/13118765
```

Screen Shot 2020-07-17 at 6 42 24 AM

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,65 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
         )
-        logger.debug(f"cmd: {cmd}")
-        with open(out, "w") as stdout, open(err, "w") as stderr:
-            retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
-        if retcode != 0:
-            raise ChildProcessError(f"Sort failed, retcode: {retcode}")
-        else:
-            shutil.move(temp_bam, bam)
-        os.remove(err)
-        os.remove(out)
```
WiscEvan, line 80: I'm not sure why we are spending time debating this. The previous implementation writes _and_ keeps the samtools.out, samtools.err files if an issue has occurred, otherwise they are removed. Your fixes should follow the same behavior.

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,43 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-    ChildProcessError
-        Samtools did not run successfully, returns the return code from subprocessing call
+    subprocess.CalledProcessError
+        Samtools did not run successfully
     """
 
-    if type(nproc) is not int or nproc <= 0:
-        raise TypeError(f"nproc must be an integer greater than zero! Given: {nproc}")
-    if not os.path.exists(sam) or os.stat(sam).st_size == 0:
+    if type(cpus) is not int or cpus <= 0:
```
WiscEvan, line 57:
```suggestion
    if not isinstance(cpus, int) or cpus <= 0:
```

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,43 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
         )
         logger.debug(f"cmd: {cmd}")
         with open(out, "w") as stdout, open(err, "w") as stderr:
-            retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
-        if retcode != 0:
-            raise ChildProcessError(f"Sort failed, retcode: {retcode}")
-        else:
-            shutil.move(temp_bam, bam)
+            subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True, check=True)
+            shutil.move(sort_fpath, bam)
```
WiscEvan, line 74: I think the previous variable name of `temp_bam` is more clear. `samtools sort` outputs a bam file. `sort_fpath` loses this detail. I think the clarity becomes more apparent at the end with `shutil.move(temp_bam, bam)` rather than `shutil.move(sort_fpath, bam)`

autometa/common/external/samtools.py
```diff
@@ -95,10 +88,15 @@ def main(args):
     parser = argparse.ArgumentParser(
         description="Takes a sam file, sorts it and returns the output to a bam file"
     )
-    parser.add_argument("--sam", help="</path/to/alignment.sam>", type=str)
-    parser.add_argument("--bam", help="</path/to/output/alignment.bam>", type=str)
+    parser.add_argument("sam", help="</path/to/alignment.sam>", type=str)
+    parser.add_argument("bam", help="</path/to/output/alignment.bam>", type=str)
     parser.add_argument(
-        "--nproc", help="Number of processors to use", default=mp.cpu_count(), type=int
+        "--cpus", help="Number of processors to use", default=mp.cpu_count(), type=int
```
WiscEvan, line 94: 👍

autometa/common/external/samtools.py
```diff
@@ -42,49 +41,41 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-            f"samtools view -@{nproc} -bS {sam} | samtools sort -@{nproc} -o {temp_bam}"
-        )
+        cmd = f"samtools view -@{cpus} -bS {sam} | samtools sort -@{cpus} -o {temp_bam}"
         logger.debug(f"cmd: {cmd}")
         with open(out, "w") as stdout, open(err, "w") as stderr:
-            retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
-        if retcode != 0:
-            raise ChildProcessError(f"Sort failed, retcode: {retcode}")
-        else:
+            subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True, check=True)
```
WiscEvan, line 71: Do you think we should wrap this in a try/except block and explicitly handle the `subprocess.CalledProcessError`?

Sidduppal: We can. But what would be the advantage to that?

autometa/common/external/samtools.py
```diff
@@ -68,7 +68,12 @@ def sort(sam, bam, cpus=mp.cpu_count()):
         cmd = f"samtools view -@{cpus} -bS {sam} | samtools sort -@{cpus} -o {temp_bam}"
         logger.debug(f"cmd: {cmd}")
         with open(out, "w") as stdout, open(err, "w") as stderr:
-            subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True, check=True)
+            proc = subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True)
+            try:
+                proc.check_returncode()
+            except subprocess.CalledProcessError as err:
+                print(f"Error while running samtools sort \n cmd: {cmd}")
+                raise err
```
WiscEvan, line 76: Couldn't you leave the check and just write the try/except around it?

```suggestion
            try:
                subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True, check=True)
            except subprocess.CalledProcessError as err:
                logger.error(f"SamtoolsSortError: {cmd}")
                raise err
```

Ultimately, since we will know where the failure occurs, we could log the error and handle the error where the function is called. We could similarly put a `SamtoolsSortError` in `exceptions.py` to handle this

autometa/common/external/samtools.py
```diff
@@ -68,7 +68,12 @@ def sort(sam, bam, cpus=mp.cpu_count()):
         cmd = f"samtools view -@{cpus} -bS {sam} | samtools sort -@{cpus} -o {temp_bam}"
         logger.debug(f"cmd: {cmd}")
         with open(out, "w") as stdout, open(err, "w") as stderr:
-            subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True, check=True)
+            proc = subprocess.run(cmd, stdout=stdout, stderr=stderr, shell=True)
+            try:
+                proc.check_returncode()
+            except subprocess.CalledProcessError as err:
+                print(f"Error while running samtools sort \n cmd: {cmd}")
```
WiscEvan: We should not be printing here. We should be using the `logger`.

```suggestion
                logger.error(f"SamtoolsSortError: {cmd}")
```

WiscEvan: Ultimately, since we will know where the failure occurs, we could log the error and handle the error where the function is called. We could similarly put a SamtoolsSortError in exceptions.py to handle this

WiscEvan:
```suggestion
from autometa.common.exceptions import SamtoolsSortError

            except subprocess.CalledProcessError as err:
                raise SamtoolsSortError(cmd, err)
```

autometa/common/exceptions.py
```diff
@@ -87,8 +87,12 @@ class SamtoolsSortError(AutometaException):
...
     """
 
-    def __init__(self, value):
-        self.value = value
+    def __init__(self, cmd, err):
+        self.cmd = cmd
+        self.err = err
+
+    def __str__(self):
+        return f"Error while running external utility \n cmd: {self.cmd} \n error: {self.err}"
```
WiscEvan, line 95: Minor nitpick. But could you swap the error and cmd display? I think more often we will want to look at the command run first before we dive into the traceback information

autometa/common/external/samtools.py
```diff
@@ -74,9 +74,8 @@ def sort(sam, bam, cpus=mp.cpu_count()):
                     cmd, stdout=stdout, stderr=stderr, shell=True, check=True
                 )
             except subprocess.CalledProcessError as err:
-                err.cmd = cmd
-                logger.error(f"SamtoolsSortError: {cmd}")
-                raise SamtoolsSortError(err)
+                logger.error(f"Error while running samtools \n cmd: {cmd}")
+                raise ExternalToolError(cmd, err)
```
WiscEvan, line 78: I think it may be more appropriate to log this exception when handling it.

```suggestion
                raise ExternalToolError(cmd, err)
```

autometa/common/coverage.py
```diff
@@ -174,8 +174,8 @@ def make_bed(lengths=lengths, fasta=fasta, bam=bam, bed=bed):
         def sort_samfile(sam=sam, bam=bam, nproc=nproc):
             try:
                 samtools.sort(sam, bam, cpus=nproc)
-            except SamtoolsSortError:
-                raise SamtoolsSortError
+            except ExternalToolError:
+                raise ExternalToolError
```
WiscEvan, line 178: Here we are handling the exception. so we may log it here. Right now you are just raising the error which is not doing anything. At the moment, there is no reason for the try/except block because the same behavior would occur without it. This is the location where we could exit gracefully while logging the error information to the initialized logger.

```suggestion
            except ExternalToolError as err:
                logger.error(err)
                # Not sure if we want to exit in this manner. But something like this...
                sys.exit(-1)
```

@jason-c-kwan may have an opinion on how we should handle exiting Autometa when an error is encountered _and handled_.

autometa/common/external/samtools.py
```diff
@@ -42,49 +42,47 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
-            retcode = subprocess.call(cmd, stdout=stdout, stderr=stderr, shell=True)
-        if retcode != 0:
-            raise ChildProcessError(f"Sort failed, retcode: {retcode}")
-        else:
+            try:
+                subprocess.run(
+                    cmd, stdout=stdout, stderr=stderr, shell=True, check=True
+                )
+            except subprocess.CalledProcessError as err:
+                logger.error(f"Error while running samtools \n cmd: {cmd}")
```
WiscEvan, line 77: I'm do not think a message should be logged since an exception is being raised.

```suggestion

```

Sidduppal:
> A few things:
> 
>     1. There is redundancy in the `ExternalToolError` exception.
> 
>     2. `coverage.py` is exception handling in the wrong location. This exception handling should be at the entrypoint.
> 
>     3. You could also do this (2.) in `samtools.py`

I moved the handling of `ExternalToolError` to `__main__.py`. However, I am unsure about the advantage of doing this...?
Can you explain what you mean by redundancy?

WiscEvan:
> > A few things:
> > ```
> > 1. There is redundancy in the `ExternalToolError` exception.
> > 
> > 2. `coverage.py` is exception handling in the wrong location. This exception handling should be at the entrypoint.
> > 
> > 3. You could also do this (2.) in `samtools.py`
> > ```
> 
> I moved the handling of `ExternalToolError` to `__main__.py`. However, I am unsure about the advantage of doing this...?
> Can you explain what you mean by redundancy in `ExternalToolError` exception?

I was referring to `main()` in `samtools.py` and `coverage.py`, not `__main__.py`.

autometa/__main__.py
```diff
@@ -179,6 +180,9 @@ def entrypoint():
 
     try:
         main(args)
+    except ExternalToolError:
+        logger.error(f"Error while running an external utility.")
+        raise ExternalToolError
```
WiscEvan, line 185: These changes should be discarded

Sidduppal:
> > > A few things:
> > > ```
> > > 1. There is redundancy in the `ExternalToolError` exception.
> > > 
> > > 2. `coverage.py` is exception handling in the wrong location. This exception handling should be at the entrypoint.
> > > 
> > > 3. You could also do this (2.) in `samtools.py`
> > > ```
> > 
> > 
> > I moved the handling of `ExternalToolError` to `__main__.py`. However, I am unsure about the advantage of doing this...?
> > Can you explain what you mean by redundancy in `ExternalToolError` exception?
> 
> I was referring to `main()` in `samtools.py` and `coverage.py`, not `__main__.py`.

I may need some help with this. This could be my lack of experience but I am not able to understand your comment. I think this may be because I am not able to understand the reasoning or purpose behind it. I looked up "exception handling at entry points" (and other similar phrases) and even went through the antismash code, but I am still not clear on our goal. Won't handling the exception in `main()` be useful only in cases where the script is run as a stand-alone module? Is it possible to cite some resources or a mini explanation?

Also when you say entry point what do you mean exactly?

autometa/common/external/samtools.py
```diff
@@ -42,49 +42,46 @@ def sort(sam, bam, nproc=mp.cpu_count()):
...
     ------
     TypeError
-        nproc must be an integer greater than zero
+        cpus must be an integer greater than zero
     FileNotFoundError
         Specified path is incorrect or the file is empty
-    ChildProcessError
-        Samtools did not run successfully, returns the return code from subprocessing call
+    subprocess.CalledProcessError
+        Samtools did not run successfully
```
WiscEvan, line 55: This is now raising an `ExternalToolError`

autometa/__main__.py
```diff
@@ -32,6 +32,7 @@
 import multiprocessing as mp
 
 from autometa.config.user import AutometaUser
+from autometa.common.exceptions import ExternalToolError
```
WiscEvan, line 35: Why is this being imported?

autometa/common/coverage.py
```diff
@@ -37,6 +37,7 @@
 from autometa.common.external import bowtie
 from autometa.common.external import samtools
 from autometa.common.external import bedtools
+from autometa.common.exceptions import ExternalToolError
```
WiscEvan, line 40: So far you are importing this but not using it?

WiscEvan:
> > > > A few things:
> > > > ```
> > > > 1. There is redundancy in the `ExternalToolError` exception.
> > > > 
> > > > 2. `coverage.py` is exception handling in the wrong location. This exception handling should be at the entrypoint.
> > > > 
> > > > 3. You could also do this (2.) in `samtools.py`
> > > > ```
> > > 
> > > 
> > > I moved the handling of `ExternalToolError` to `__main__.py`. However, I am unsure about the advantage of doing this...?
> > > Can you explain what you mean by redundancy in `ExternalToolError` exception?
> > 
> > 
> > I was referring to `main()` in `samtools.py` and `coverage.py`, not `__main__.py`.
> 
> I may need some help with this. This could be my lack of experience but I am not able to understand your comment. I think this may be because I am not able to understand the reasoning or purpose behind it. I looked up "exception handling at entry points" (and other similar phrases) and even went through the antismash code, but I am still not clear on our goal. Won't handling the exception in `main()` be useful only in cases where the script is run as a stand-alone module? Is it possible to cite some resources or a mini explanation?
> 
> Also when you say entry point what do you mean exactly?

Placing a try/except around an entrypoint (i.e. `autometa.common.coverage.main()`) allows handling of unhandled exceptions. This provides the opportunity for us to emit the traceback to the log and exit gracefully. 

See AntiSMASH example:

https://github.com/antismash/antismash/blob/6215cd84d16a229eaba36759c1f556c006b8f45c/antismash/__main__.py#L114-L122

>Won't handling the exception in `main()` be useful only in cases where the script is run as a stand-alone module?

Precisely. `main()`, in this scenario, is the entrypoint to the `coverage.py` stand-alone module.

Some references:

1. [good practice to have an unhandled exception handler at the highest level of the call stack](https://stackoverflow.com/a/2737337/13118765)
3. [Python entrypoints](https://stackoverflow.com/a/782984/13118765)
4. [Another discussion on when to write try/except and how to handle exceptions](https://softwareengineering.stackexchange.com/a/97893)

Sidduppal:
> > > > > A few things:
> > > > > ```
> > > > > 1. There is redundancy in the `ExternalToolError` exception.
> > > > > 
> > > > > 2. `coverage.py` is exception handling in the wrong location. This exception handling should be at the entrypoint.
> > > > > 
> > > > > 3. You could also do this (2.) in `samtools.py`
> > > > > ```
> > > > 
> > > > 
> > > > I moved the handling of `ExternalToolError` to `__main__.py`. However, I am unsure about the advantage of doing this...?
> > > > Can you explain what you mean by redundancy in `ExternalToolError` exception?
> > > 
> > > 
> > > I was referring to `main()` in `samtools.py` and `coverage.py`, not `__main__.py`.
> > 
> > 
> > I may need some help with this. This could be my lack of experience but I am not able to understand your comment. I think this may be because I am not able to understand the reasoning or purpose behind it. I looked up "exception handling at entry points" (and other similar phrases) and even went through the antismash code, but I am still not clear on our goal. Won't handling the exception in `main()` be useful only in cases where the script is run as a stand-alone module? Is it possible to cite some resources or a mini explanation?
> > Also when you say entry point what do you mean exactly?
> 
> Placing a try/except around an entrypoint (i.e. `autometa.common.coverage:main`) allows handling of unhandled exceptions. This provides the opportunity for us to emit the traceback to the log and exit gracefully.
> 
> See AntiSMASH example:
> https://github.com/antismash/antismash/blob/6215cd84d16a229eaba36759c1f556c006b8f45c/antismash/__main__.py#L114-L122
> 
> Entrypoints are typically defined in the `setup.py` file:
> 
> https://github.com/KwanLab/Autometa/blob/aaa7ff5c435be39fdaf95fe7c401e6f8a1789ee8/setup.py#L24-L32
> 
> > Won't handling the exception in `main()` be useful only in cases where the script is run as a stand-alone module?
> 
> Precisely. `main()`, in this scenario, is the entrypoint to the `coverage.py` stand-alone module.
> 
> Some references:
> 
>     1. [good practice to have an unhandled exception handler at the highest level of the call stack](https://stackoverflow.com/a/2737337/13118765)
> 
>     2. [Python entrypoints](https://stackoverflow.com/a/782984/13118765)
> 
>     3. [Another discussion on when to write try/except and how to handle exceptions](https://softwareengineering.stackexchange.com/a/97893)

As per this I think we have three options then.
1. We can either handle the `ExternalToolError` at the entrypoint of `coverage.py` or `samtools.py` ie. at  `autometa.common.coverage:main` or  `autometa.common.samtools:main` but that would only be useful when the script is run as a stand-alone module (which would be less often) than when it is run as a part of the autometa pipeline (which would be more often the case)
2. Handle it at the point where `sort` is being called in `coverage.py`
here:
https://github.com/KwanLab/Autometa/blob/aaa7ff5c435be39fdaf95fe7c401e6f8a1789ee8/autometa/common/coverage.py#L173-L174
3. Handle it at the highest call stack ie. at the `main()` function of `__main__.py` file.
here:
https://github.com/KwanLab/Autometa/blob/aaa7ff5c435be39fdaf95fe7c401e6f8a1789ee8/autometa/__main__.py#L180-L195
I personally would like option 3 as that is the highest call stack. 
Feel free to let me know in case I have missed something :smile:

autometa/__main__.py
WiscEvan: This needs to be removed from `__main__.py`
```suggestion

```

autometa/common/coverage.py
WiscEvan: Since you are not using this here, you should remove it from the file.

autometa/common/external/samtools.py
WiscEvan: This needs to be updated to `ExternalToolError`

```suggestion
    ExternalToolError
        Samtools did not run successfully, returns subprocess traceback and command run
```

WiscEvan:
> As per this I think we have three options then.
> 
> 1. We can either handle the `ExternalToolError` at the entrypoint of `coverage.py` or `samtools.py` ie. at  `autometa.common.coverage:main` or  `autometa.common.samtools:main` but that would only be useful when the script is run as a stand-alone module (which would be less often) than when it is run as a part of the autometa pipeline (which would be more often the case)
> 2. Handle it at the point where `sort` is being called in `coverage.py`
>    here:
>    https://github.com/KwanLab/Autometa/blob/aaa7ff5c435be39fdaf95fe7c401e6f8a1789ee8/autometa/common/coverage.py#L173-L174
> 3. Handle it at the highest call stack ie. at the `main()` function of `__main__.py` file.
>    here:
>    https://github.com/KwanLab/Autometa/blob/aaa7ff5c435be39fdaf95fe7c401e6f8a1789ee8/autometa/__main__.py#L180-L195
>    
>    I personally would like option 3 as that is the highest call stack.
>    Feel free to let me know in case I have missed something 😄


Putting `ExternalToolError` in `__main__.py` is not appropriate in this context. We are already catching _any_ exception with `except Exception as err`. I am referring to putting a `try/except` block in the main functions of `coverage.py` When running `autometa-coverage` we would want an `ExternalToolError` to be raised and for Autometa to exit gracefully. So far, this is _not_ in place. 


Something like this would capture the error messages for debugging while also not flooding the user's screen with traceback information.

```python
try:
    main()
except ExternalToolError as err:
    logger.error(err)
    sys.exit(-1)
```

You could construct a few extra attributes that you think would be most helpful for debugging and for guidance to the user when initializing the `ExternalToolError` exception.

i.e.

```python
try:
    main()
except ExternalToolError as err:
    logger.error(err)
    logger.info(f"{err.tool} failed at command: {err.command}")
    sys.exit(-1)
```

autometa/common/coverage.py
WiscEvan, line 40:
```suggestion

```

autometa/common/external/samtools.py
```diff
@@ -95,10 +92,15 @@ def main(args):
...
-    parser.add_argument("--bam", help="</path/to/output/alignment.bam>", type=str)
+    parser.add_argument("sam", help="</path/to/alignment.sam>", type=str)
+    parser.add_argument("bam", help="</path/to/output/alignment.bam>", type=str)
     parser.add_argument(
-        "--nproc", help="Number of processors to use", default=mp.cpu_count(), type=int
+        "--cpus", help="Number of processors to use", default=mp.cpu_count(), type=int
     )
     args = parser.parse_args()
-    main(args)
+    sort(args.sam, args.bam, args.cpus)
```
WiscEvan, line 101:
```suggestion
    sort(sam=args.sam, bam=args.bam, cpus=args.cpus)
```

Issue #3, opened 2020-02-06
Update Docker image and wrappers


WiscEvan:
- dev branch Dockerfile
- run script for user running docker container

jason-c-kwan:
Also - make sure that dockerhub is building the dev branch Dockerfile when made, and that the resulting container works. Note - Autometa 1 had wrapper scripts to make it easy for workflows to be run through docker easily. Do we need to implement those?

WiscEvan:
We should be able to install autometa within the docker container with the same entrypoints as you would have access to if you installed autometa using conda. We could write one wrapper script to handle all of these entrypoints within the container. That will likely be the easiest route.

jason-c-kwan:
I've updated Docker hub so that it also builds a 'dev' image. This should help in testing the dev dockerfile.

jason-c-kwan:
OK, so the dev build had errors, I think stemming from there not being a python3 'tsne' module found (didn't you make one @WiscEvan ? Anyway, here is the log:

```
Cloning into '.'...
Warning: Permanently added the RSA host key for IP address '140.82.112.3' to the list of known hosts.
Switched to a new branch 'dev'
KernelVersion: 4.4.0-1060-aws
Components: [{u'Version': u'19.03.8', u'Name': u'Engine', u'Details': {u'KernelVersion': u'4.4.0-1060-aws', u'Os': u'linux', u'BuildTime': u'2020-03-11T01:24:30.000000000+00:00', u'ApiVersion': u'1.40', u'MinAPIVersion': u'1.12', u'GitCommit': u'afacb8b7f0', u'Arch': u'amd64', u'Experimental': u'false', u'GoVersion': u'go1.12.17'}}, {u'Version': u'1.2.13', u'Name': u'containerd', u'Details': {u'GitCommit': u'7ad184331fa3e55e52b890ea95e65ba581ae3429'}}, {u'Version': u'1.0.0-rc10', u'Name': u'runc', u'Details': {u'GitCommit': u'dc9208a3303feef5b3839f4323d9beb36df0a9dd'}}, {u'Version': u'0.18.0', u'Name': u'docker-init', u'Details': {u'GitCommit': u'fec3683'}}]
Arch: amd64
BuildTime: 2020-03-11T01:24:30.000000000+00:00
ApiVersion: 1.40
Platform: {u'Name': u'Docker Engine - Community'}
Version: 19.03.8
MinAPIVersion: 1.12
GitCommit: afacb8b7f0
[...]

PR #123, opened 2021-03-10
Fix Dockerfile

WiscEvan:
- :art::bug::green_heart::penguin: Fix Dockerfile. Now capable of running autometa entrypoints within docker container
- Fixes issue #3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM continuumio/miniconda3
-MAINTAINER Jason C. Kwan "jason.kwan@wisc.edu"
+LABEL maintainer="jason.kwan@wisc.edu"
 
-# Copyright 2020 Ian J. Miller, Evan R. Rees, Kyle Wolf, Siddharth Uppal,
+# Copyright 2021 Ian J. Miller, Evan R. Rees, Kyle Wolf, Siddharth Uppal,
 # Shaurya Chanana, Izaak Miller, Jason C. Kwan
 #
 # This file is part of Autometa.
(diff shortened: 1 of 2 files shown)


chasemc (commented):
Consider adding at least minimum versions for the Conda install libraries

Dockerfile
```diff
@@ -19,15 +19,45 @@ MAINTAINER Jason C. Kwan "jason.kwan@wisc.edu"
...
+RUN autometa-length-filter -h \
+    & autometa-orfs -h \
+    & autometa-coverage -h \
+    & autometa-kmers -h \
+    & autometa-markers -h \
+    & autometa-taxonomy -h \
+    & autometa-taxonomy-lca -h \
+    & autometa-taxonomy-majority-vote -h \
+    & autometa-binning -h \
+    & autometa-unclustered-recruitment -h
```
jason-c-kwan, line 63: The output of this only shows up when you build the docker image, not when you run it. So no one will see this. Perhaps put this in the bash rc file?

jason-c-kwan (changes requested):
Dockerfile builds and runs on my local machine. One issue noted to do with the help message only showing up during build and not run. Please note: I did not do extensive testing of the pipeline in the docker image. The Docker hub build will only really be tested when this is merged (due to the Docker hub settings), but I predict it will at least build.

Dockerfile
WiscEvan: Do you want the help text for all of the entrypoints to be printed to the user each time a container is spawned from the image? This is a minimal image as it does not have any of the required databases configured. The user will need to mount their databases to the docker container and then have to point autometa to these mounted paths. That being said, I think nextflow can be configured such that the user will not have to worry about issuing these docker commands. We can just hardcode the docker image into each process and nextflow will do the rest.

jason-c-kwan: Oh, I see. It looked like your intention was to show the messages to the user. In that case, you could probably take them out after showing it works. On another note though, if the Docker is intended to be used interactively, then a message upon startup might be a good idea.

jason-c-kwan (approved):
Looks good, the image builds on my local machine.

PR #129, opened 2021-03-12
Add support for gzipped assemblies

WiscEvan:
:art: resolves KwanLab/Autometa/issue#125

### DIFF
--- a/autometa/common/coverage.py
+++ b/autometa/common/coverage.py
@@ -25,6 +25,7 @@ Calculates coverage of contigs
 """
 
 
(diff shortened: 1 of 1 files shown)


WiscEvan:
Well the code coverage report does not seem to make sense to me. There is lower coverage for `unclustered_recruitment.py`, however the only file that has changed is `autometa/common/coverage.py`. I think this can safely be ignored for the merge

jason-c-kwan (approved):
Looks good!

PR #131, opened 2021-03-12
Update bug report template (master branch)

WiscEvan:
Now asks users to paste contents of run_autometa.log and provide any additional information that they can muster up themselves (utilities to retrieve this information using autometa are not available in this branch).


jason-c-kwan (approved):
Looks great!

Issue #110, opened 2020-09-23
API for loading or fetching data for development, benchmarking and testing


WiscEvan:
Fetch or load data similar to the [sklearn API](https://scikit-learn.org/stable/datasets/index.html). This would provide a number of utilities to load, fetch and/or download toy and real world datasets. Each dataset that is loaded via the API should first pass a file integrity check by comparing checksums. Some example load/fetch/download functions may include:

|                  function | dataset                                         | paper                           |
|--------------------------:|-------------------------------------------------|---------------------------------|
| `load_random(...)`        | load random number of genomes fetched from NCBI | N/A                             |
| `load_simulated(...)`     | load Autometa simulated community               | DOI: 10.1093/nar/gkz148         |
| `load_sharon(...)`        | load Sharon dataset                             | DOI: 10.1101/gr.142315.112      |
| `load_synthetic(...)`     | load Autometa synthetic community               | DOI: 10.1093/nar/gkz148         |
| `load_environmental(...)` | load Autometa environmental community           | DOI: 10.1093/nar/gkz148         |
| `load_cami(...)`          | load CAMI challenge dataset                     | DOI: 10.1186/s12859-020-03667-3 |

Example usage could look something like this:

```python
# import load function
from autometa.datasets import load_synthetic
# load in to data object
data = load_synthetic("MIX51")
```

data object in this case should also contain some metadata (similar to sklearn API)

WiscEvan:
We have uploaded some test datasets to a public shared google drive. The links are below:

simulated communities: [drive folder](https://drive.google.com/open?id=1JFjVb-pfQTv4GXqvqRuTOZTfKdT0MwhN "simulated communities google drive folder")
synthetic communitites: [drive folder](https://drive.google.com/open?id=1ez0QBzOsPSmwT4aNFCnshADFkK7FOF4t "synthetic communities google drive folder")

WiscEvan:
A simple helper script could be written that uses gdown to download to the user provided output directory.

A simple example:

```python
import gdown
# ... 
# ...
# NOTE: dictionary of communities with their respective metagenome.fna.gz file IDs
simulated = {
    "78": "15CB8rmQaHTGy7gWtZedfBJkrwr51bb2y"
}
file_id = ...
url = f"https://drive.google.com/uc?id={file_id}"
gdown.download(url, args.output)
```

I've skipped a few details above as this may be nice practice for someone 

👀 @ajlail98 

[...]

PR #121, opened 2021-03-06
Update documentation

Sidduppal:
:memo: Add step-by-step tutorial on how to run Autometa    
:fire: Remove Rest API
:art: Add docs/source/_build to .gitignore
:memo: Update autometa install guidelines. Added docker to it.
:memo: Add benchmarking page
:memo: Add Automappa to examining results
:art: Replaced shell with bash in parse_argparse.py
:memo: Add packages to install for developers in contributing guidelines"

PR #138, opened 2021-03-25
Add feature to download google drive datasets

ajlail98:
Issue #110


chasemc:
Don't worry about `nextflow`, but `gdown` must be added to `requirements.txt`
```
PackagesNotFoundError: The following packages are not available from current channels:

  - gdown
  - nextflow
 ```

**Comment by WiscEvan** [comment]
`gdown` is already in `requirements.txt` ([here](https://github.com/KwanLab/Autometa/blob/dev/requirements.txt#L2)). Try `git pull upstream dev` to update your branch with the most recent commits on dev. After you have resolved any merge conflicts, then push again 👍

**Review by WiscEvan — COMMENTED** [review]
This looks like a great start. To get an idea on the formatting for constructing this as a proper module of Autometa, see the template script in the Autometa/docs folder ([🔗   template.py](https://github.com/KwanLab/Autometa/blob/dev/docs/template.py))

**Comment by chasemc** [comment]
> `gdown` is already in `requirements.txt` ([here](https://github.com/KwanLab/Autometa/blob/dev/requirements.txt#L2)). Try `git pull upstream dev` to update your branch with the most recent commits on dev. After you have resolved any merge conflicts, then push again 👍
> 
> This is also apparent from the 18 files that would be changed by this pull request

That's why I deleted the comment. For some reason make wasn't  building but maybe that'll be fixed after pulling from upstream or it was just my system

File: autometa/validation/download_dataset.py [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+    'test': '1fy3M7RnS_HGSQVKidCy-rAwXuxldyOOv',
+    '78': '15CB8rmQaHTGy7gWtZedfBJkrwr51bb2y',
+    '156': '13bkwFBIUhdWVWlAmVCimDODWF-7tRxgI',
+    '312': '1qyAu-m6NCNuVlDFFC10waOD28j15yfV-',
+    '625': '1FgMXSD50ggu0UJbZd1PM_AvLt-E7gJix',
+    '1250': '1KoxwxBAYcz8Xz9H2v17N9CHOZ-WXWS5m',
+    '2500': '1wKZytjC4zjTuhHdNUyAT6wVbuDDIwk2m',
+    '5000': '1IX6vLfBptPxhL44dLa6jePs-GRw2XJ3S',
+    '10000': '1ON2vxEWC5FHyyPqlfZ0znMgnQ1fTirqG'
+}
```
Sidduppal, line 40: Looks like a great start. You can try the `items()` method to loop through the dictionary and get the respective "ID" needed to download.

WiscEvan:
Just noting here (requesting some additional functionality that would be really convenient for benchmarking studies 😉 )

Ability to download specific file from a dataset:

e.g. Just a thought on the usage. I'm not sure how you'd like to implement this in practice.

```bash
autometa-download-dataset \
    --community 78 \
    --file metagenome \
    --output $HOME/78Mbp.metagenome.fna.gz
```

This would find the respective community, get the file that was specified (metagenome.fna.gz from metagenome) and download it to the path provided from `--output`. 

There may be a more intuitive way to set up the arguments, but I have run into a few instances where I'd like to only download one or a few files and would not like the entire 78 community dataset directory.

Dockerfile
```diff
@@ -46,4 +46,4 @@ RUN echo "Checking autometa entrypoints" \
     && autometa-markers -h > /dev/null \
     && autometa-taxonomy -h > /dev/null \
     && autometa-binning -h > /dev/null \
-    && autometa-unclustered-recruitment -h > /dev/null 
+    && autometa-unclustered-recruitment -h > /dev/null
```
chasemc, line 49: OK. Only change is auto-formatting

autometa/binning/recursive_dbscan.py
```diff
@@ -931,7 +931,10 @@ def main():
         default="bacteria",
     )
```
chasemc, line 932: OK. Only change is auto-formatting

autometa/binning/unclustered_recruitment.py
```diff
@@ -400,7 +400,9 @@ def get_confidence_filtered_predictions(
         raise NotImplementedError(classifier)
 
     df = pd.DataFrame(
-        predictions, index=test_data.index, columns=train_data.target_names,
+        predictions,
+        index=test_data.index,
+        columns=train_data.target_names,
     )
     # Filter predictions by confidence threshold
```
chasemc, line 407: ✅ Only change is auto-formatting

autometa/common/kmers.py
```diff
@@ -673,7 +673,10 @@ def main():
         default=False,
     )
     parser.add_argument(
-        "--cpus", help=f"num. processors to use.", default=cpus, type=int,
+        "--cpus",
+        help=f"num. processors to use.",
+        default=cpus,
```
chasemc, line 678: ✅ Only change is auto-formatting

autometa/binning/summary.py
```diff
@@ -303,7 +303,8 @@ def main():
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
     parser.add_argument(
-        "workspace", help="Path to Autometa results workspace directory",
+        "workspace",
+        help="Path to Autometa results workspace directory",
```
chasemc, line 307: ✅ Only change is auto-formatting

autometa/common/markers.py
```diff
@@ -170,7 +170,11 @@ def get(
 
     if not os.path.exists(out) or not os.path.getsize(out):
```
chasemc, line 171: ✅ Only change is auto-formatting

autometa/common/metagenome.py
```diff
@@ -471,7 +472,10 @@ def main():
     args = parser.parse_args()
     dirpath = os.path.dirname(os.path.realpath(args.assembly))
     raw_mg = Metagenome(
-        assembly=args.assembly, outdir=dirpath, prot_orfs_fpath="", nucl_orfs_fpath="",
+        assembly=args.assembly,
+        outdir=dirpath,
+        prot_orfs_fpath="",
+        nucl_orfs_fpath="",
     )
```
chasemc, line 480: ✅ Only change is auto-formatting

autometa/config/environ.py
```diff
@@ -237,7 +237,9 @@ def get_versions(program=None):
         try:
             return globals()[exe_name]()
         except TypeError:
-            logger.warning(f"{exe_name} not found. This may impact a stage of the Autometa pipeline.")
+            logger.warning(
+                f"{exe_name} not found. This may impact a stage of the Autometa pipeline."
+            )
             return "Not found"
     versions = {}
```
chasemc, line 244: ✅ Only change is auto-formatting

autometa/config/databases.py
```diff
@@ -769,7 +775,10 @@ def main():
 
     config = get_config(args.config)
     dbs = Databases(
-        config=config, dryrun=args.dryrun, nproc=args.nproc, update=args.update,
+        config=config,
+        dryrun=args.dryrun,
+        nproc=args.nproc,
+        update=args.update,
     )
```
chasemc, line 782: ✅ Only change is auto-formatting

autometa/config/project.py
```diff
@@ -130,8 +130,7 @@ def new_metagenome_num(self):
             return mg_num
 
     def save(self):
-        """Save project config in project directory
-        """
+        """Save project config in project directory"""
         put_config(self.config, self.config_fpath)
 
     def new_metagenome_directory(self):
```
chasemc, line 136: ✅ Only change is auto-formatting

autometa/taxonomy/majority_vote.py
```diff
@@ -298,7 +298,12 @@ def majority_vote(
         filename, __ = os.path.splitext(os.path.basename(orfs))
         outdir = os.path.dirname(os.path.realpath(out))
         lca_out = os.path.join(outdir, ".".join([filename, "lca.tsv"]))
-    lca_fpath = lca.blast2lca(orfs=orfs, out=lca_out, blast=blast, force=force,)
+    lca_fpath = lca.blast2lca(
+        orfs=orfs,
+        out=lca_out,
+        blast=blast,
+        force=force,
```
chasemc, line 305: ✅ Only change is auto-formatting

tests/unit_tests/test_samtools.py
```diff
@@ -32,6 +32,7 @@ def test_sort_missing_file():
     with pytest.raises(FileNotFoundError):
         samtools.sort(sam="sam", bam="bam")
```
chasemc, line 34: ✅ Only change is auto-formatting

tests/unit_tests/test_summary.py
```diff
@@ -202,7 +202,8 @@ def return_metabin_taxonomies(*args, **kwargs):
 
 @pytest.mark.skip
 def test_get_metabin_taxonomies(
-    mock_rank_taxids, bin_df,
+    mock_rank_taxids,
+    bin_df,
 ):
     mock_ncbi = return_mock_ncbi()
```
chasemc, line 208: ✅ Only change is auto-formatting

tests/unit_tests/test_unclustered_recruitment.py
```diff
@@ -113,7 +113,8 @@ def test_get_kmer_features(kmers, dimensions):
 
 
 def test_get_features_no_taxa(
-    kmers, coverage,
+    kmers,
+    coverage,
 ):
```
chasemc, line 118: ✅ Only change is auto-formatting

tests/unit_tests/test_vote.py
```diff
@@ -191,7 +210,10 @@ def test_write_ranks_no_taxonomy_columns(tmp_path, votes):
     assembly = dirpath / "assembly.fna"
     with pytest.raises(KeyError):
         vote.write_ranks(
-            taxonomy=votes, assembly=assembly, outdir=dirpath, rank="superkingdom",
+            taxonomy=votes,
+            assembly=assembly,
+            outdir=dirpath,
+            rank="superkingdom",
         )
```
chasemc, line 218: ✅ Only change is auto-formatting

autometa/config/project.py
```diff
@@ -0,0 +1,325 @@
+#!/usr/bin/env python
```
jason-c-kwan, line 1: This file does not appear to exist in the latest commit in `dev`, so you should delete the file here to avoid re-adding it upon merge.

autometa/validation/download_dataset.py
```diff
@@ -0,0 +1,47 @@
+# in response to issue #110
```
jason-c-kwan, line 1: Please include the copyright notice that is found at the top of other scripts.

autometa/validation/download_dataset.py
```diff
@@ -0,0 +1,47 @@
...
+    '5000': '1IX6vLfBptPxhL44dLa6jePs-GRw2XJ3S',
+    '10000': '1ON2vxEWC5FHyyPqlfZ0znMgnQ1fTirqG'
+}
+
+# construct file id into a url to put into gdown
+file_id = simulated[args.community]
+url = f'https://drive.google.com/uc?id={file_id}'
+
+# download the specified file with gdown
+gdown.download(url, args.output)
```
jason-c-kwan, line 47: Something is going wrong for me running this script. I tried:

```bash
python download_dataset.py --community 78 --output ~/Downloads/test_data_download
```

And it downloaded a single file called:

```
test_data_downloadexk955shtmp
```

I don't know what this file is, but running `unzip` gives an error, and `tar xvf` seems to run but not do anything.

autometa/validation/download_dataset.py
```diff
@@ -0,0 +1,47 @@
+# in response to issue #110
+# Intended: fetch data similarly to scikit-learn API
```
jason-c-kwan, line 2: Please make this file executable with `chmod +x autometa/validation/download_dataset.py`, then do a commit.

jason-c-kwan (changes requested):
This does generally look good. I don't know whether you intend on adding the functionality that @WiscEvan requested, but I found some things that probably should be addressed for the merge. I don't think these will be too hard to do.

1. The `autometa/config/project.py` file doesn't exist in the `dev` branch any more, so you should delete it before the merge. 
2. The `download_dataset.py` file should have the standard copyright notice and the script should be made executable (you can look up the chmod command in linux/macos for more information).
3. I encountered a problem when running the script on macos (see comment). I'm honestly not sure what the problem is, so @WiscEvan might be able to help you with troubleshooting.

autometa/config/project.py
jason-c-kwan: This could also be solved with an upstream pull from `dev`, so you could try that first because it will be easier.

autometa/validation/download_dataset.py
chasemc: I had no idea git preserved permissions

WiscEvan:
It appears you do not have pre-commit hooks setup nor am I able to find a section on this in the documentation (PR #121). Maybe I'm not looking hard enough? @Sidduppal

Sidduppal:
@WiscEvan installation of pre-commit hooks is mentioned under Contributing guidelines on GitHub ([link](https://github.com/KwanLab/Autometa/blob/dev/.github/CONTRIBUTING.md)). The link to that is in the documentation. Do you want me to copyover them to the documentation as well?

autometa/validation/download_dataset.py
ajlail98: I couldn't figure out what that file is or what went wrong. When Sidd or I run it now, it works for the 78 MBP dataset. Let me know if this problem is persisting on your setup.

autometa/common/external/download_dataset.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+import gdown
+import os
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def download_dataset(dataset, out_dirpath):
```
WiscEvan, line 35: Minor nitpick but could this be renamed to `download` since `dataset` is implied? I've also suggested using [type hints](https://docs.python.org/3/library/typing.html). This is not necessary but it helps when developing in VSCode or some other text editor later on.

```suggestion
def download(dataset: str, out_dirpath: str) -> None:
```

autometa/common/external/download_dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        "312": "1qyAu-m6NCNuVlDFFC10waOD28j15yfV-",
+        "625": "1FgMXSD50ggu0UJbZd1PM_AvLt-E7gJix",
+        "1250": "1KoxwxBAYcz8Xz9H2v17N9CHOZ-WXWS5m",
+        "2500": "1wKZytjC4zjTuhHdNUyAT6wVbuDDIwk2m",
+        "5000": "1IX6vLfBptPxhL44dLa6jePs-GRw2XJ3S",
+        "10000": "1ON2vxEWC5FHyyPqlfZ0znMgnQ1fTirqG",
+    }
+
+    # construct file id into a url to put into gdown
+    file_id = simulated[dataset]
```
WiscEvan, line 50: What happens if I use `download(dataset=9999, out_dirpath="$HOME/benchmarking_data"`? A check for the provided value should be put in place and a `KeyError` or `ValueError` raised if the provided value is incorrect.

```suggestion
    if dataset not in simulated:
        raise KeyError(f"dataset ({dataset}) not in simulated data. Choices: {simulated.keys()}")
    file_id = simulated[dataset]
```

autometa/common/external/download_dataset.py
```diff
@@ -0,0 +1,90 @@
...
+    url = f"https://drive.google.com/uc?id={file_id}"
+    filename = f"{dataset}_metagenome.fna.gz"
+    out_fpath = os.path.join(out_dirpath, filename)
```
WiscEvan, line 53: Currently this only allows us to download each dataset's metagenome fasta file. Can we add functionality here to download any of the available files for a particular dataset? In terms of benchmarking, this functionality would be really helpful/useful. For example, if we are testing a number of different parameter sets for binning, it would be convenient to benchmark these against the existing binning file from Autometa v1.0 (available on google drive) as well as retrieve the ground-truth assignments (also available on google drive) to determine benchmarking metrics.

I've briefly described some example usage in the benchmarking pull request (https://github.com/KwanLab/Autometa/pull/141#issuecomment-847348201) detailing the usage within a nextflow process. Can we implement this functionality?

### Suggested change

Change the arguments to the function definition

```python
# From
def download(dataset: str, out_dirpath: str):
    ...
# To
def download(dataset: str, file: str, out: str) -> str:
    # 1. lookup dataset
    # 2. lookup file in dataset
    # 3. Write file to out
    ...
    return out
```

autometa/common/external/download_dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        "test": "1fy3M7RnS_HGSQVKidCy-rAwXuxldyOOv",
+        "78": "15CB8rmQaHTGy7gWtZedfBJkrwr51bb2y",
+        "156": "13bkwFBIUhdWVWlAmVCimDODWF-7tRxgI",
+        "312": "1qyAu-m6NCNuVlDFFC10waOD28j15yfV-",
+        "625": "1FgMXSD50ggu0UJbZd1PM_AvLt-E7gJix",
+        "1250": "1KoxwxBAYcz8Xz9H2v17N9CHOZ-WXWS5m",
+        "2500": "1wKZytjC4zjTuhHdNUyAT6wVbuDDIwk2m",
+        "5000": "1IX6vLfBptPxhL44dLa6jePs-GRw2XJ3S",
+        "10000": "1ON2vxEWC5FHyyPqlfZ0znMgnQ1fTirqG",
+    }
```
WiscEvan, line 47: Can you separate simulated from test data? This could become unwieldy moving forward with other datasets.

autometa/common/external/download_dataset.py
WiscEvan, line 35: This function needs a docstring

autometa/common/external/download_dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        required=True,
+    )
+    parser.add_argument(
+        "--out_dirpath",
+        help="specify the directory to download the file",
+        required=True,
+    )
+    args = parser.parse_args()
+
+    download_dataset(args.dataset, args.out_dirpath)
```
WiscEvan, line 86: With some of the other suggested changes, this main logic will have to be altered. It would be nice if a user could provide multiple input arguments to dataset, the file from the dataset as well as the corresponding output file path.

In this case you will need to change the `parser.add_argument(...)` method slightly (see [nargs](https://docs.python.org/3/library/argparse.html#nargs) keyword argument of `parser.add_argument(...)`) as well as change the way you are calling `download(...)`. What would be particularly helpful in getting `dataset`, `file` and `out` to all correspond to their respective inputs would be the python built-in [zip  function](https://docs.python.org/3.8/library/functions.html#zip). This could be done in main to call all of the retrieved inputs appropriately with `download`


I'll leave some details for you to figure out, but the idea is here.

```python
# Example use of zip with download call and retrieving corresponding function arguments
for dataset, filetype, filepath in zip(datasets, filetypes, filepaths):
    download(dataset=dataset, file=filetype, out=filepath)
```

WiscEvan (changes requested):
This is nice progress 👍 🔧 👷 

I've placed comments throughout and made suggestions where appropriate. Also, a minor nitpick but can you move this file to validation and rename it?

e.g.

```
Autometa/autometa/validation/datasets.py
```

This way when we want to implement some of this down the line within some autometa benchmarking/validation script, we can write

```python
# import
from autometa.validation import datasets
# call function from module
datasets.download(...)
```

Instead of 

```python
# import
from autometa.validation import download_datasets
# call function from module
download_datasets.download(...)
```

Let me know if you have any questions and/or disagree with some of my suggestions! Happy to hear your thoughts! 🧠 🤯 😄

jason-c-kwan:
I won't do a full review right now because I guess @WiscEvan has some things to look into, but I just tried this and it seems to work on my Mac now. Not sure what the problem was. Perhaps some Google Drive weirdness.

Sidduppal:
@WiscEvan @ajlail98 
> I've placed comments throughout and made suggestions where appropriate. Also, a minor nitpick but can you move this file to validation and rename it?

Asking andrew to move it to external directory was my fault. I was under the assumption that the validation directory is for version 1 autometa scripts only and since we are using an "external" tool it should be under external directory. But now I think about it your argument makes much more sense.

autometa/common/external/download_dataset.py
ajlail98: Good call, it is done.

autometa/common/external/download_dataset.py
ajlail98: I've just added new ability to download entire folders, specific files, or all files with the same name across folders. I haven't figured out Nextflow enough to say whether or not this will play nice there. I'm trying to just do one thing at a time! Does it look a bit more useful now?

autometa/common/external/download_dataset.py
ajlail98: Removed test data, now I just test using README.md.

autometa/validation/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+    key_list = [*targets_dict["file"]]
+    for key in key_list:
+        # retrieve values from targets_dict
+        file = targets_dict["file"][key]
+        folder = targets_dict["folder"][key]
+        file_id = targets_dict["file_id"][key]
+
+        # construct file id into a url to put into gdown
+        url = f"https://drive.google.com/uc?id={file_id}"
+        filename = f"{folder}_{file}"
```
WiscEvan, line 98: I'm a little confused by this line. What is the intended behavior?

It appears the directory's and file's names are being joined together into one string rather than placing the file in the provided `folder` directory. I think the most intuitive arguments for the `download` function would be

```python
def download(dataset:str, file: str, out:str)->None:
```

This will require two look-ups:

1. Look for `dataset` of available datasets (could be a datasets dict)
2. Look for `file` in dataset (could be a dict corresponding to respective dataset)

The corresponding `file_id` should be retrieved after successful lookup of 1 and 2.

Now with the `file_id` you can continue as you have done by constructing the url and performing the gdown download. In this case the gdown call uses the provided `out` variable.

e.g.

```python
gdown.download(url, out)
```

By the way, does `gdown.download(...)` return anything? A response code perhaps? If so, could you also return this and handle appropriately? One example would be if an error response is returned, this could be propagated back to the end-user.

autometa/validation/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+    import logging as logger
+
+    logger.basicConfig(
+        format="[%(asctime)s %(levelname)s] %(name)s: %(message)s",
+        datefmt="%m/%d/%Y %I:%M:%S %p",
+        level=logger.DEBUG,
+    )
+
+    parser = argparse.ArgumentParser(
+        prog="autometa-download-dataset",
```
WiscEvan, line 118: The entrypoint should be specified in `setup.py`

```suggestion
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+            "78.125Mbp",
+            "156.25Mbp",
+            "312.5Mbp",
+            "625Mbp",
+            "1250Mbp",
+            "2500Mbp",
+            "5000Mbp",
+            "10000Mbp",
+        ],
+    )
```
WiscEvan, line 134: This might be nitpicky, but does `dataset` seem more appropriate?

`autometa-download --dataset 78.125Mbp`

v.s.

`autometa-download --folder 78.125Mbp`

Also, I think this should be a required argument where the user must explicitly specify a dataset. Perhaps another choice here could be `"all"` where the user has to explicitly state that they would like to download all of the datasets. This way you could add the extra keyword argument to the `add_argument(...)` method

```suggestion
        "--dataset",
        help="specify a simulated community size to download",
        choices=[
            "78.125Mbp",
            "156.25Mbp",
            "312.5Mbp",
            "625Mbp",
            "1250Mbp",
            "2500Mbp",
            "5000Mbp",
            "10000Mbp",
            "all"
        ],
        required=True,
    )
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+            "kmers.tsv.gz",
+            "markers.tsv.gz",
+            "Bacteria.fna.gz",
+            "orfs.faa.gz",
+            "metagenome.filtered.fna.gz",
+            "hmmscan.tsv.gz",
+            "forward_reads.fastq.gz",
+            "reverse_reads.fastq.gz",
+        ],
+    )
```
WiscEvan, line 162:
```suggestion
            "reverse_reads.fastq.gz",
        ],
        default="all",
    )
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+    parser.add_argument(
+        "--out_dirpath",
```
WiscEvan, line 164: Minor nitpick here

```suggestion
        "--output",
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+        help="specify the directory to download the file(s)",
+        required=True,
+    )
+    args = parser.parse_args()
+
+    download(fetchIDs(args.folder, args.file), args.out_dirpath)
```
WiscEvan, line 170: I have fallen into a similar trap here in the past. Consider splitting up some of your main logic. Ultimately it should be readable (_almost like pseudocode_) in describing its functionality. You've got some of the logic implemented into your functions, but now that they're nested, it has become a bit more difficult to digest. See below for an example:

Example bash call:

```bash
autometa-datasets \
    --community 78 \
    --file metagenome,reference_assignments,binning \
    --output $HOME/78Mbp.metagenome.fna.gz,$HOME/78Mbp.assignments.tsv.gz,$HOME/78Mbp.binning.tsv.gz
```

NOTE: It may be easiest to consider only allowing the user to specify one community. This would play nicely with nextflow and simplify your task. Another thing to notice from the above example is the number of arguments supplied to `--file` and `--output` are the same. They correspond to each other such that we can look them up in our python code.

e.g.

```python
def main():
    ############
    args = parser.parse_args()
    dataset = args.community
    filetypes = args.file
    filepaths = args.output
    ############
    # NOTE: You will need to look up some argparse docs for the above to properly work (see `nargs`)
    for filetype, filepath in zip(filetypes, filepaths):
        # The following behavior is more intuitive. It almost reads out as a sentence
        download(dataset=dataset, file=filetype, out=filepath)
        # "download" the "dataset" of the "filetype" to "filepath"
```

You could similarly use your `fetchIDs(...)` function, but I would break this up in the `main` logic.


```python
file_ids = fetchIDs(...)
for file_id in file_ids:
    # Notice we have no method here to specify
    # the output file path which is behavior that 
    # would be convenient for the end-user.
    download(..., file_id, ...)
```

WiscEvan (changes requested):
Interesting changes 🧵  it appears you may have wound yourself into a knot 🥨 . I think you could simplify this to make it more convenient to the devs and the end-user. In regards to nextflow, at the most basic level we just want to download a file (or three) at a time but _all corresponding to the same community_ so the feature of downloading all of the simulated communities at once is a _nice to have_ but doesn't necessarily help us. Otherwise, it may be important to consider the amount of data that would be downloaded if the user does not provide any arguments. Would a user like this as intended behavior? If so or if we're unsure, should we emit a warning or confirmation before downloading all of this data? All things to ponder, but not necessary to implement. Just food for thought. 🌭 

Anyways, it looks like you've gone down the rabbit hole 🐰 🕳️ . I think if we rewind and focus on the simple case we shouldn't be too far off. Overall though, it looks like it is coming along nicely 👍 . Iterate, iterate, iterate... ⚙️  and so it goes. Keep up the good work 🏗️ 🚧 👷 👏

WiscEvan:
Function to check internet connection from Autometa repo:

[internet_is_connected function](https://github.com/KwanLab/Autometa/blob/50f7a60c329acfa9e705b4cf343dfdcc21ade0ec/autometa/config/databases.py#L186-L195)

autometa/validation/datasets.py
ajlail98: I went ahead and removed the "all" functionality for now, but I'll see about putting it back in later on.

autometa/validation/datasets.py
ajlail98: I don't think gdown returns anything.

autometa/validation/datasets.py
ajlail98: Now you can select one folder, multiple files, and the main logic is cleaner. Thanks for the suggestions!

autometa/validation/datasets.py
ajlail98: This should be good now? I'd appreciate you giving it a look to make sure I didn't break anything. First time under the hood and all...

autometa/validation/datasets.py
```diff
@@ -0,0 +1,212 @@
...
+import sys
+import logging
+import pandas as pd
+
+from autometa.common.utilities import internet_is_connected
+
+logger = logging.getLogger(__name__)
+
+
+def download(type: str, size: list, file_names: list, dir_path: str) -> None:
```
chasemc, line 39: Preference thing about style but keeping consistent variable names will help others understand your code and also you when you come back to it later. 
e.g. instead of shortening 'community_type' to 'type' within "download()" just keep it "community_type". I tend to err on the side of my variable names being too descriptive
     download(type=community_type, size=community_size, file_names=file_names, dir_path=dir_path)

autometa/validation/datasets.py
```diff
@@ -0,0 +1,212 @@
...
+
+    """
+
+    if type == "synthetic" or type == "all":
+        sys.exit("Haven't implemented that yet")
+
+    for community_size in size:
+        index = pd.read_csv("gdown_fileIDs.csv", dtype=str)
+        index = index.query(f'dataset == "{community_size}"')
+        dir_path_size = ""
```
chasemc, line 67:
```suggestion
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,212 @@
...
+        dir_path_size = os.path.join(dir_path, community_size)
+        # make a new directory
+        if not os.path.exists(dir_path_size):
```
chasemc, line 70: Need to make sure the provided `dir_path` exists or mkdir will fail

autometa/validation/datasets.py
```diff
@@ -0,0 +1,212 @@
...
+            os.mkdir(dir_path_size)
+
+        for file_name in file_names:
+            file_id = ""
```
chasemc, line 74:
```suggestion
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,212 @@
...
+            file_id = index.query(f'file == "{file_name}"')["file_id"].to_list()
+            file_id = file_id[0]
+            dir_path_final = ""
```
chasemc, line 77:
```suggestion
```

chasemc (commented):
Internet slow today so didn't test functionality

Couple of notes:
Variable initialization isn't always necessary and can be overridden unlike in some other languages

```
>>> a = ""
>>> a = 1
>>> type(a)
<class 'int'>
>>> 
>>> del a
>>> 
>>> a + ""
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'a' is not defined
>>> 
>>> 
>>> a = 1
>>> 
>>> a + ""
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'int' and 'str'
>>> 
>>> 
>>> a = ""
>>> 
>>> a + "+ string"
'+ string'
```


I would avoid naming your pandas table `index` as index is also pandas method  
i.e.
```
>>> index = pd.read_csv("/home/chase/Documents/github/Autometa/autometa/validation/gdown_fileIDs.csv", dtype=str)
>>> index.index
RangeIndex(start=0, stop=167, step=1)
```
If not using a descriptive variable name most people use `df`

autometa/common/utilities.py
```diff
@@ -444,6 +445,18 @@ def wrapper(*args, **kwds):
     return wrapper
 
 
+def internet_is_connected(
+    host: str = "8.8.8.8", port: int = 53, timeout: int = 2
```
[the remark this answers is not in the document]
ajlail98: I've written my error warning such that everything keeps going as normal, so it wouldn't be affected by that kind of issue. But should we change this from google to another site? Might be something to talk to @WiscEvan about.

autometa/validation/datasets.py
ajlail98: Good catch! I added a logger error in main().

setup.py
```diff
@@ -37,6 +37,7 @@ def read(fname):
             "autometa-binning = autometa.binning.recursive_dbscan:main",
             "autometa-unclustered-recruitment = autometa.binning.unclustered_recruitment:main",
             "autometa-binning-summary = autometa.binning.summary:main",
+            "autometa-download-dataset = autometa.validation.datasets:main",
```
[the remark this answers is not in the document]
ajlail98: I added one? I think that's how it's supposed to go, assuming there's no reason to avoid having one there.

autometa/common/external/download_dataset.py
ajlail98: These days I'm using argparser, so this should be automatically handled to return an error.

autometa/common/utilities.py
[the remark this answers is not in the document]
chasemc: If you aren't going to fail at that point then you could just add to the warning something like: "... internet check may incorrectly fail where google.com is not reachable/ping-able"

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+
+    Returns
+    -------
+    None
+        download is completed through gdown
+
+    """
+
+    if community_type == "synthetic" or community_type == "all":
+        sys.exit("Haven't implemented that yet")
```
WiscEvan, line 64: Python has built-in exceptions for this case

```suggestion
        raise NotImplemented
```

```suggestion
        raise NotImplementedError
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+        # make a new directory
+        if not os.path.exists(dir_path_size):
+            os.mkdir(dir_path_size)
+
+        for file_name in file_names:
+            file_id = df.query(f'file == "{file_name}"')["file_id"].to_list()[0]
+            dir_path_final = os.path.join(dir_path_size, file_name)
+            url = f"https://drive.google.com/uc?id={file_id}"
+
+            gdown.download(url, dir_path_final)
```
WiscEvan, line 79: `dir_path_size` and `dir_path_final` correspond to the `community_size` output directory and the `community_size` file, respectively. 

- (minor nitpick) Could you rename these variables to represent the directory and file?

```suggestion
        community_size_outdir = os.path.join(dir_path, each_community_size)
        # make a new directory
        if not os.path.exists(community_size_outdir):
            os.makedirs(community_size_outdir)

        for file_name in file_names:
            file_id = df.set_index(["dataset", "file"]).loc[(community_size, file_name), 'file_id']
            community_size_filepath = os.path.join(community_size_outdir, file_name)
            url = f"https://drive.google.com/uc?id={file_id}"

            gdown.download(url, community_size_filepath)
```

- (minor nitpick) Similar to the query you performed above, could the quoting usage be reduced?

I've noticed when I have to look-up a particular value in pandas, indexing is one of the first things that comes to mind. Indexing can be quite powerful to access a particular value for multiple columns.

## Example:

```python
# hierarchical index when reading in the table
df = pd.read_csv("gdown_fileIDs.csv", dtype=str, index_col=["dataset", "file"])
```

### Setting the index after we've read in the table.

```python
## Two methods: 
### 1. in place, no variable re-assignment (will return NoneType)
df.set_index(['dataset', 'file'], inplace=True)
### 2. with variable re-assignment (returns the dataframe with the set index)
df = df.set_index(['dataset', 'file'])
```

### Looking up a value with the hierarchical index

```python
# Make sure hierarchical index is set
df = df.set_index(['dataset', 'file'])
community_size = '78Mbp'
file_name = 'orfs.faa.gz'
df.loc[(community_size, file_name), 'file_id']
# The same can all be done without variable reassignment or having a new df with the set index.
### input
df.set_index(['dataset', 'file']).loc[('78Mbp', 'orfs.faa.gz'), 'file_id']
### output
'1z-VveBEeVRyr7eushHvHCrdqDShNkaBq'
```

### Looking up all values with hierarchical index

Look-up for a generic pandas Dataframe usage: `df.loc[index, cols]`

```python
# Make sure hierarchical index is set
df = df.set_index(['dataset', 'file'])
community_size = '78Mbp'
file_name = 'orfs.faa.gz'
# In
df.loc[(community_size, file_name)] # Note there is no ',' after the 2-tuple.
# Out
community                            simulated
file_id      1z-VveBEeVRyr7eushHvHCrdqDShNkaBq
Name: (78Mbp, orfs.faa.gz), dtype: object

## NOTE: the columns can be accessed via there namespace with this as well
df.loc[(community_size, file_name)].file_id
df.loc[(community_size, file_name)].community
```

## Original

```python
file_id = df.query(f'file == "{file_name}"')["file_id"].to_list()[0]
````
## Other suggestions using index

```python
df = df.set_index(['community', 'file'])
# Access within loc (IMO preferred)
file_id = df.loc[(community_size, file_name), "file_id"]
# Access via namespace
file_id = df.loc[(community_size, file_name)].file_id
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+    import logging as logger
+
+    logger.basicConfig(
+        format="[%(asctime)s %(levelname)s] %(name)s: %(message)s",
+        datefmt="%m/%d/%Y %I:%M:%S %p",
+        level=logger.DEBUG,
+    )
+
+    parser = argparse.ArgumentParser(
+        description="Download a simulated community file from google drive to a specified dir_path_size"
```
WiscEvan, line 93:
```suggestion
        description="Download a simulated community file from google drive to a specified output directory"
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+    parser.add_argument(
+        "--community_type",
```
WiscEvan, line 96: - (minor nitpick) replace underscore with hyphen for reduced keystrokes (Accessed the same with argparse)

```suggestion
        "--community-type",
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+        help="specify synthetic or simulated communities (currently only simulated is available)",
+        choices=[
+            "synthetic",
+            "simulated",
+            "all",
+        ],
+        required=True,
+    )
+    parser.add_argument(
+        "--community_size",
```
WiscEvan, line 106: (minor nitpick) replace underscore with hyphen for reduced keystrokes (Accessed the same with argparse)

```suggestion
        "--community-size",
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+            "2500Mbp",
+            "5000Mbp",
+            "10000Mbp",
+            "all",
+        ],
+        required=True,
+        nargs="+",
+    )
+    parser.add_argument(
+        "--file_names",
```
WiscEvan, line 123: - (minor nitpick) replace underscore with hyphen for reduced keystrokes (Accessed the same with argparse)

```suggestion
        "--file-names",
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+            "hmmscan.tsv.gz",
+            "forward_reads.fastq.gz",
+            "reverse_reads.fastq.gz",
+            "all",
+        ],
+        nargs="+",
+        required=True,
+    )
+    parser.add_argument(
+        "--dir_path",
```
WiscEvan, line 153: - (minor nitpick) replace underscore with hyphen for reduced keystrokes (Accessed the same with argparse)

```suggestion
        "--dir-path",
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+        help="specify a folder to start the download (several directories will be generated within this folder)",
+        required=True,
+    )
+    args = parser.parse_args()
+
+    community_type = args.community_type
```
WiscEvan, line 159: - :fire: reassignment of the variable is not necessary here. You can just pass in `args.community_type` to `download(...)`

```suggestion
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+            "156Mbp",
+            "312Mbp",
+            "625Mbp",
+            "1250Mbp",
+            "2500Mbp",
+            "5000Mbp",
+            "10000Mbp",
+        )
+    else:
+        community_size = args.community_size
```
WiscEvan, line 172: This should work for now, but as we add datasets, it may be more appropriate to retrieve all communities from the gdown file_ids table instead of hardcoding them in here. 

```suggestion
    if "all" in args.community_size:
        community_sizes = (
            "78Mbp",
            "156Mbp",
            "312Mbp",
            "625Mbp",
            "1250Mbp",
            "2500Mbp",
            "5000Mbp",
            "10000Mbp",
        )
    else:
        community_sizes = args.community_size
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+            "Bacteria.fna.gz",
+            "orfs.faa.gz",
+            "metagenome.filtered.fna.gz",
+            "hmmscan.tsv.gz",
+            "forward_reads.fastq.gz",
+            "reverse_reads.fastq.gz",
+        )
+    else:
+        file_names = args.file_names
+    dir_path = args.dir_path
```
WiscEvan, line 199: :fire: Variable reassignment is not necessary here.

```suggestion
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+
+    if not internet_is_connected():
```
WiscEvan, line 201: I believe there is an argument for this function that you could pass through argparse. e.g. The ability to ping a different address.

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+        logger.error(
+            "No internet connection detected. Please confirm connection. Downloader will still attempt to run. (Internet check may incorrectly fail where google.com is not reachable/ping-able (e.g. China))"
+        )
+
+    if not os.path.exists(dir_path):
+        logger.error(
+            "The dir_path you specified could not be found. Please select a folder that exists."
```
WiscEvan, line 208:
```suggestion
    if not os.path.exists(args.dir_path):
        logger.error(
            f"{args.dir_path} specified could not be found. Please select a folder that exists."
```

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        sys.exit("Specified path does not exist!")
```
WiscEvan, line 210:
```suggestion
        sys.exit("{args.dir_path} does not exist!")
```

autometa/validation/datasets.py
WiscEvan, line 208: The `download(...)` function can be implemented to create the root output directory as well as sub-directories. If this becomes the case, these checks may be removed.

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+
+    download(
+        type=community_type,
+        size=community_size,
+        file_names=file_names,
+        dir_path=dir_path,
```
WiscEvan, line 216:
```suggestion
        type=args.community_type,
        size=community_sizes,
        file_names=file_names,
        dir_path=args.dir_path,
```

WiscEvan (changes requested):
A few suggestions, requested changes and minor nitpicks. Looking pretty good. 👷 🔧

autometa/validation/datasets.py
```diff
@@ -0,0 +1,221 @@
...
+        download is completed through gdown
+
+    """
+
+    if community_type == "synthetic" or community_type == "all":
+        sys.exit("Haven't implemented that yet")
+
+    for each_community_size in community_size:
+        df = pd.read_csv("gdown_fileIDs.csv", dtype=str)
+        df = df.query(f'dataset == "{each_community_size}"')
```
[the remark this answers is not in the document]
ajlail98: Just tried it without the double quotes enclosing the `{community_size}`, and that doesn't work. I'll move to the `community_df = df[df.dataset == community_size]` format instead.

autometa/validation/datasets.py
ajlail98: Fantastic! This is so much more elegant than how I was doing it. Thanks for the guidance!

autometa/validation/datasets.py
ajlail98: Removed them!

autometa/validation/datasets.py
ajlail98: I'm thinking it's not worth the extra argparse complexity for including this. If the check fails, the program still tries to run, so I think it's probably best to leave it as a "heads-up" warning.

autometa/validation/datasets.py
ajlail98: I agree that I should revisit this in the future, but I know it would take me a long time to figure out how to do it dynamically, so it's probably a problem for later. There will be more added complexity when I add in the synthetic datasets, anyway.

autometa/validation/datasets.py
WiscEvan: Perhaps. However, how is the program supposed to behave if the internet is not connected? Similarly, adding a parameter to the entrypoint would provide a solution for the edge case Chase mentioned while also helping to account for when the internet is truly not connected, versus just trying to connect to a banned IP address.


```python
## Addition of optional host parameter
parser.add_argumet(
    "--host",
    help="IP address to ping when checking internet connectivity. Note: Will attempt to connect to port 43 on host address (Default is google.com)",
    default="8.8.8.8"
)
...

if not internet_is_connected(host=args.host):
    # Now (assuming the user is aware of their IP bans and has input an appropriate host address) 
    # we know the user is not connected to the internet and can appropriately modify the programs behavior.
    ...

```

ajlail98: Ok, I've added that argument.

autometa/validation/datasets.py
```diff
@@ -0,0 +1,217 @@
...
+    -------
+    None
+        download is completed through gdown
+
+    """
+
+    if community_type == "synthetic" or community_type == "all":
+        raise NotImplementedError
+
+    df = pd.read_csv("gdown_fileIDs.csv", dtype=str, index_col=["dataset", "file"])
```
WiscEvan, line 66: I'm not sure whether this should be hardcoded here. If this is hardcoded here, would it be worth specifying the github URL or google drive URL for this file instead of the file on the user's system? 

1. If the script has made it to this stage, it should be able to connect.
2. This should resolve any issues of the script not being able to find the appropriate path to this file.

WiscEvan (approved):
Looks good to me. I just made one comment. Curious to hear your thoughts 👍 🎉

autometa/validation/datasets.py
ajlail98: Oh, I didn't even test that! Good catch. I think you're right, it should redirect out somewhere we know the file is. Do you think github or google drive would be more appropriate? I am thinking google drive would be best since the url would never have to change. It should be a small enough file that we can just use pd.read_csv() directly, without any gdown. (However, I don't have permission to add things to the simulated communities folder where I think this file should be kept.)

autometa/validation/gdown_fileIDs.csv
```diff
@@ -0,0 +1,168 @@
+community,dataset,file,file_id
```
WiscEvan, line 1: We should be able to remove this from the PR now that it is hosted on google drive. Correct?

ajlail98: Yep!

autometa/validation/datasets.py
```diff
@@ -0,0 +1,223 @@
...
+            "hmmscan.tsv.gz",
+            "forward_reads.fastq.gz",
+            "reverse_reads.fastq.gz",
+        )
+    else:
+        file_names = args.file_names
+
+    if not internet_is_connected(host=args.host):
+        logger.error(
+            "No internet connection detected. Please confirm connection. Downloader will still attempt to run. (Ping a custom IP address with --host argument)"
```
jason-c-kwan, line 211:
```suggestion
            "No internet connection detected (couldn't ping google.com at IP 8.8.8.8). Please confirm connection. Downloader will still attempt to run. (Ping a custom IP address with --host argument)"
```

jason-c-kwan: This is a bit more explanatory. A user might not know what pinging is or why it is relevant to this error.

setup.py
jason-c-kwan, line 40: You need to add this endpoint to the end of the `Dockerfile`, where you have a test of the other endpoints.

jason-c-kwan (changes requested):
Overall looks good, I just had some minor comments. You need to add your endpoint to the `Dockerfile` that is used when building a Docker image. I also suggested you change the error message for no internet to add context for the user. Optional, but I suggest you can try the Chinese railway website (`114.114.114.114`) if `8.8.8.8` fails.

jason-c-kwan (approved):
Looks great! Well done :tada:

PR #139, opened 2021-04-08
only try to generate hmm indexes if they do not exist

hyphaltip:
This fix is one in a step to supporting global system install on HPCC with read-only access to marker folder


jason-c-kwan (approved):
This looks great. Thanks for contributing!

PR #148, opened 2021-04-26
Only ignore /lib at top level

chasemc:
The nf-core template has a `lib` directory.  This commit changes .gitignore to only ignore `lib` at the top level directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
+/lib/
 lib64/
 parts/
 sdist/


WiscEvan:
Where is `nf-core` respective to this? It is hard to tell why this change is necessary from the current changes.

Does this mean there is a `lib` sub-directory of the `nextflow` directory?

chasemc:
> Does this mean there is a `lib` sub-directory of the `nextflow` directory?

Yes.  `Autometa/nextflow/lib/`

chasemc:
Converted to issue 
https://github.com/KwanLab/Autometa/issues/156

Issue #170, opened 2021-06-03
Write pickle files in defined location


chasemc:
Yes, during the `lca.py` step a few different data structures are constructed from the NCBI database files. These are used to quickly look-up LCA values from the precomputed sparse arrray of `precomputed_lcas.pkl.gz`. 

## Definition of serialized files

These filepaths are defined here:

https://github.com/KwanLab/Autometa/blob/50f7a60c329acfa9e705b4cf343dfdcc21ade0ec/autometa/taxonomy/lca.py#L102-L108

## Datastructure preparation 

These are prepared with the method `prepare_lca`

https://github.com/KwanLab/Autometa/blob/50f7a60c329acfa9e705b4cf343dfdcc21ade0ec/autometa/taxonomy/lca.py#L276-L291

## Use/Look-up of serialized files

https://github.com/KwanLab/Autometa/blob/50f7a60c329acfa9e705b4cf343dfdcc21ade0ec/autometa/taxonomy/lca.py#L293-L347

## Additional Notes

Here is a reference from [topcoder](https://www.topcoder.com/thrive/articles/Range%20Minimum%20Query%20and%20Lowest%20Common%20Ancestor "topcoder Range Minimum Query and Lowest Common Ancestor Article") for a better explanation of this algorithm

### Tree of Life Creation

The tree of life is constructed from the node.dmp file from NCBI's taxonomy database located in the taxdump.tar.gz compressed file. Branches stemming from the root are constructed until the entire tree has been built. Paths between each node by are built by traversing the tree using an Eulerian tour method. During the Eulerian tour, features of each tax ID are stored for sparse table creation.

### Sparse Table

Sparse Table: The depth of each taxonomic ID in relation to the rest of the tree is used to efficiently store the tree in memory for quick lookup of taxonomic information. This algorithm employs dynamic programming to assess each tax ID in a range of other tax IDs starting from a range of the tax ID and it's closest relative and increasing to a range from the tax ID and it's furthest relative.

### RMQ

[...]

PR #141, opened 2021-04-19
Classification and Clustering Benchmarking

WiscEvan:
## Overview

1. :art: Add `autometa-benchmark` entrypoint under `autometa.validation.benchmark` module
2. :art::green_apple: Add `BENCHMARK_CLUSTERING` nextflow process
3. :art::green_apple: Add `BENCHMARK_CLASSIFICATION` nextflow process
4. :art::green_apple: Add `publishDir` operator to `BINNING_SUMMARY` process.

### `autometa-benchmark` entrypoint

Add `autometa-benchmark` entrypoint for classification and clustering benchmarking against autometa reference communities. This adds functionality to easily compute classification and clustering benchmarking statistics of test datasets against the current 'gold' standard (Set as Autometa v.1 predictions).

### `BENCHMARK_CLUSTERING` nextflow process

~**NOTE: PR #138 entrypoint required for Nextflow process**~

This process takes as input all binning/clustering predictions as well as the `community` value associated with these binning predictions. ~This requires the to-be-added `autometa-datasets` entrypoint from PR #138.~ Two tables are written (wide and long format) with clustering metrics computed using scikit-learn's clustering metrics.

### Example Figure: Simulated Community Clustering Metrics

These metrics were plotted using the long-formatted (`--output-long`) output file generated by `autometa-benchmark` with `--benchmark=clustering`

simulated_community_clustering_evaluation_metrics

### `BENCHMARK_CLASSIFICATION` nextflow process

~**NOTE: PR #138 entrypoint required for Nextflow process**~

This process takes as input all taxonomic predictions (tab-delimited file of contig and taxid columns) as well as the `community` value associated with these taxonomic predictions. ~This requires the to-be-added `autometa-datasets` entrypoint from PR #138.~ One table is written with overall classification metrics computed using scikit-learn's classification metrics. Additionally a classification report is written per community. Metrics are computed using taxids such that all taxonomic predictions may be classified (not only canonical-ranks). This allows assessment of taxonomic predictions grouped by canonical and non-canonical ranks.

### Example Figure: Median precision vs. recall of taxonomic predictions grouped by non-canonical ranks. 

These metrics were plotted using the output file generated by `autometa-benchmark` with `--benchmark=classification`

**Figures are faceted by simulated community dataset**

support values are the number of taxa within each taxonomic label class

median_scores_by_rank

PR #182, opened 2021-08-03
Large data mode

WiscEvan:
## Behavior

Iterates through taxa and selects contigs to cluster based on a provided upper bound (`--max-partition-size`). If the taxon set is below the upper bound, the taxon's k-mer counts are normalized and embedded prior to recursive clustering. If the taxon set is below the lower bound (`min_contigs = max([pca_dimensions + 1, embed_dimensions + 1])`) k-mer's are retrieved from the canonical rank embedding then the same recursive clustering implementation begins.

## Features/Additions

* New entrypoint: `autometa-large-data-mode-binning`
  * `--max-partition-size`
  * `--cache`
  * `--binning-checkpoints`
* Add binning checkpoint behavior
* Add embedding caching behavior
* Add new entrypoint `autometa-large-data-mode-binning-loginfo` that will parse log file generated from `autometa-large-data-mode-binning`
  * Differentiate timing between normalization, embedding, misc. and clustering
* Rename entrypoint `autometa-parse-bed` to `autometa-bedtools-genomecov`


chasemc:
I made some edits to make the binning code faster within the nfcore branch, and you've obviously made a way more edits here. It would be really nice to be able to run something like you created in https://github.com/KwanLab/Autometa/pull/141 to check these PRs for regression/changes in the binning perform. Maybe that PR review should be prioritized?

autometa/taxonomy/lca.py
```diff
@@ -94,18 +92,19 @@ class LCA(NCBI):
...
+        self.tour_fp = os.path.join(self.outdir, "tour.pkl.gz")
         self.tour = None
-        self.level_fp = os.path.join(self.dbdir, "level.pkl.gz")
+        self.level_fp = os.path.join(self.outdir, "level.pkl.gz")
         self.level = None
-        self.occurrence_fp = os.path.join(self.dbdir, "occurrence.pkl.gz")
+        self.occurrence_fp = os.path.join(self.outdir, "occurrence.pkl.gz")
         self.occurrence = None
-        self.sparse_fp = os.path.join(self.dbdir, "precomputed_lcas.pkl.gz")
+        self.sparse_fp = os.path.join(self.outdir, "precomputed_lcas.pkl.gz")
```
WiscEvan, line 107: This should resolve #170

autometa/taxonomy/lca.py
```diff
@@ -94,18 +92,19 @@ class LCA(NCBI):
 
     """
 
-    def __init__(self, dbdir: str, verbose: bool = False):
+    def __init__(self, dbdir: str, verbose: bool = False, outdir: str = None):
```
WiscEvan, line 95: Now pickled database files will be written for each metagenome to their respective output directory. This should resolve #170

chasemc: That would mean it would have to process/pickle the databases for every input?
It might be beneficial to run as a separate process in Nextflow so that [storeDir](https://www.nextflow.io/docs/latest/process.html#storedir) could be used?

chasemc:
@WiscEvan  Can you run the `autometa-benchmarks` entrypoint on this branch vs. dev and attach the results? 
I think that's going to be a good thing to do on future PR's that may affect results

PR #189, opened 2021-09-24
📝  Modify documentation for benchmark.rst

Sidduppal:
📝 Add docs for classification and clustering-classification
📝 Add docs to run multiple results of the same community at once
🔥 Remove code to aggregate results


docs/source/benchmarking.rst
```diff
@@ -13,66 +13,10 @@ Benchmarking
...
-    import pandas as pd
-    import os
-    import glob
-    dfs = []
-    for fp in glob.glob("*.clustering_benchmarks.long.tsv.gz"):
-        df = pd.read_csv(fp, sep="\t", index_col="dataset")
-        df.index = df.index.map(lambda fpath: os.path.basename(fpath))
-        dfs.append(df)
-    df = pd.concat(dfs)
-    df.to_csv("benchmarks.tsv", sep='\t', index=True, header=True)
```
WiscEvan, line 72: I don't think the code to aggregate benchmarking results should be removed from the documentation. This is a reference for anyone wanting to be able to quickly merge all of their benchmarking results.

Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN conda install -c bioconda -c conda-forge --file=requirements.txt \
     && python -m pip install trimap
 
 
+
```
WiscEvan, line 33: I'm not sure this change is necessary

docs/source/benchmarking.rst
Sidduppal: I understand. 
I perharps was confused between the two cases. I ran both of the code chunks (when index is unique and not unique) for `625Mbp.clustering_benchmarks.long.tsv.gz` and `78Mbp.clustering_benchmarks.long.tsv.gz` and git the same results.
Can you please elaborate on the differences between the two and when the user should chose which.

docs/source/benchmarking.rst
```diff
@@ -127,11 +77,92 @@ Example for the 78Mbp simulated community
...
 
-    Unfortunately, at the moment ``gdown`` doesn't support downloading entire directories from Google drive.
+    2. Unfortunately, at the moment ``gdown`` doesn't support downloading entire directories from Google drive.
     There is an open `Pull request <https://github.com/wkentaro/gdown/pull/90#issue-569060398>`_ on the ``gdown`` repository
     addressing this specific issue which we are keeping a close eye on and will update this documentation when it is merged.
 
+Benchmark clustering
+^^^^^^^^^^^^^^^^^^^^
+
+Here we are benchmarking the clustering step. All the simulated communities are being benchmarked. You don't have to run all of them, this is just for tutorial purposes.
```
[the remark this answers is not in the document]
WiscEvan: I think this simply needs to show usage rather than be a full-fledged tutorial. If that is something we would like to do, we should do it in the [KwanLab/metaBenchmarks](https://github.com/KwanLab/metaBenchmarks "https://github.com/KwanLab/metaBenchmarks") repo

docs/source/benchmarking.rst
```diff
@@ -127,11 +77,92 @@ Example for the 78Mbp simulated community
...
+        --benchmark classification \
+        --predictions $taxonomy ${community_size}.taxonomy.tsv.gz \
+        --reference simulated/${community_size}/reference_assignments.tsv.gz \
+        --output-wide ${community_size}.classification_benchmarks.wide.tsv.gz \
+        --output-classification-reports taxa_reports \
+        --ncbi /ncbi
+
+Benchmark clustering-classification
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+This will benchmark the taxonomic assignemnt of each bin. To run this you'll need ``binning.tsv.gz`` file for the specific community. This can be downloa by running:
```
WiscEvan, line 120: This does not benchmark the taxonomic assignment of each bin. This determines precision, recall and F1 score (classification metrics) on the final resulting binning

```suggestion

This will benchmark the final binning results using the classification metrics, 
precision, recall and F1-score.

.. note::

    First, F1-scores are determined for each reference genome across all recovered 
    genome bins (specifically for clusters containing any of the reference genome's 
    contigs). Then, each reference genome's F1-score is selected corresponding 
    to the respective cluster with the largest set of the respective reference genome's 
    contigs (by length in bp).
```

docs/source/benchmarking.rst
WiscEvan: If you inspect the python code, you will see the generated benchmarks are being read into pandas dataframes and the index is being set to "dataset". e.g. `pd.read_csv(..., index_col="dataset")`. You will also notice that if you know the "dataset" values are non-unique for your generated benchmarking files (you can find this out through manual inspection or by handling with python/R/bash) then the index (dataset value) will be renamed to the filename of the respective file. This was mentioned in https://github.com/KwanLab/Autometa/pull/141#issuecomment-847348201 so it is placed here as convenience for anyone who may need to aggregate multiple binning results (whether they have unique dataset values or not).

docs/source/benchmarking.rst
```diff
@@ -127,11 +77,92 @@ Example for the 78Mbp simulated community
...
+.. code:: bash
+
+    autometa-benchmark \
+        --benchmark binning-classification \
+        --predictions 78Mbp/binning.tsv \
+        --reference 78Mbp/reference_assignments.tsv.gz \
+        --output-wide binning_classification_wide.tsv.gz
+
+Specifying mulile results
+^^^^^^^^^^^^^^^^^^^^^^^^^
```
WiscEvan, line 140: I don't think specifying multiple `binning`/`classification`/`binning-classification` results needs an entire section. An example of this already exists above for benchmarking `--classification`. I think you could delete this :fire:

docs/source/benchmarking.rst
```diff
@@ -127,11 +77,92 @@ Example for the 78Mbp simulated community
...
+
+``autometa-benchmark`` provides the flexibility of specifying multiple output files of the same dataset and benchmarking them simultaneously. For example, in case you have done the clustering for a sample using three different binner, you can benchmark all of them simulateously. This is really handy when benchmarking multiple tools on the same dataset.
```
WiscEvan, line 142: ❗  Could you place this with the autometa-clustering example in the tutorial and add it under a note?❗ 

```suggestion

.. note::

    ``autometa-benchmark`` provides the flexibility of benchmarking multiple results 
    corresponding to the same dataset. For example, if you have clustering results for 
    a sample using three different binners, you can benchmark all of them, simultaneously. 
    This is really handy when benchmarking multiple tools on the same dataset.
```

WiscEvan (changes requested):
It seems the general usage and tutorial is a bit mixed and may be misleading. 

Could you address why you have rearranged/removed/added some of the sections?

I've placed suggested changes and comments throughout. 👍

docs/source/benchmarking.rst
```diff
@@ -13,66 +13,10 @@ Benchmarking
...
-        df.index = df.index.map(lambda fpath: os.path.basename(fpath))
-        dfs.append(df)
-    df = pd.concat(dfs)
-    df.to_csv("benchmarks.tsv", sep='\t', index=True, header=True)
-
 Downloading Test Datasets
-=========================
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The first step in benchmarking is to download the test data files. We will be benchmarking the taxon-profiling step and binning steps and thus the files for the same will be downloaded. reference_assignments are needed as somthiing to compare out output with. Same reference_assignmentsfile will be used for both benchmarking.
```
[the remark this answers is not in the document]
Sidduppal: What you are saying makes total sense.
However, whenever I use any documentation, I see it as a good practice when the developers mention what particular command is doing and why I'm doing it.  I have made some changes, let me know what you think.

docs/source/benchmarking.rst
Sidduppal: I understand what you mean and it that makes sense.
However, I personally feel that a usage of a command is more clear if the developers give a clear example or two.

docs/source/benchmarking.rst
```diff
@@ -42,8 +52,15 @@ Benchmark all of the simulated communities and their reference assignments
             --output-long ${community_size}.clustering_benchmarks.long.tsv.gz
     done
 
-Aggregate across simulated communities (when dataset index is unique)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Aggregate resulst across simulated communities
```
WiscEvan, line 55: This seems less descriptive to me than the previous header. Each of these header sections will be indexed along the side-bar in readthedocs. In this case, I think more descriptive section headers would be easier to quickly look-up

docs/source/benchmarking.rst
```diff
@@ -55,8 +72,10 @@ Aggregate across simulated communities (when dataset index is unique)
     ])
     df.to_csv("benchmarks.tsv", sep='\t', index=True, header=True)
 
-Aggregate across simulated communities (when dataset index is `not` unique)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When dataset index is `not` unique
```
WiscEvan, line 75: This seems less descriptive to me than the previous header. Each of these header sections will be indexed along the side-bar in readthedocs. In this case, I think more descriptive section headers would be easier to quickly look-up

docs/source/benchmarking.rst
```diff
@@ -71,73 +90,84 @@ Aggregate across simulated communities (when dataset index is `not` unique)
...
-    # Note: community is the test dataset that was used for clustering or classification. e.g.
-    # choices: 78Mbp,156Mbp,312Mbp,625Mbp,1250Mbp,2500Mbp,5000Mbp,10000Mbp
-    community_sizes=(78Mbp 156Mbp 312Mbp 625Mbp 1250Mbp 2500Mbp 5000Mbp 10000Mbp)
-
-    autometa-download-dataset \
-    --community-type simulated \
-    --community-sizes ${community_sizes[@]} \
-    --file-names reference_assignments.tsv.gz binning.tsv.gz taxonomy.tsv.gz \
-    --dir-path simulated
+    autometa-benchmark \
```
WiscEvan, line 100: Unassigned variables should not be in the documentation. `${community_size}` should be assigned in the same code block.

docs/source/benchmarking.rst
```diff
@@ -71,73 +90,84 @@ Aggregate across simulated communities (when dataset index is `not` unique)
...
----------------------------------
+This will benchmark the final binning results using the classification metrics, 
+precision, recall and F1-score.
 
-You can download the individual assemblies of different datasests with the help of ``gdown`` using command line.
-If you have installed ``autometa`` using ``conda`` then ``gdown`` should already be installed.
-If not, you can install it using ``conda install -c conda-forge gdown`` or ``pip install gdown``.
+.. note::
+    
+    First, F1-scores are determined for each reference genome across all recovered genome bins (specifically for clusters containing any of the reference genome's contigs). Then, each reference genome's F1-score is selected corresponding to the respective cluster with the largest set of the respective reference genome's contigs (by length in bp).
```
WiscEvan, line 116: Can you add newlines here? Having all of the text on one line is difficult to read in a text editor.

docs/source/benchmarking.rst
```diff
@@ -71,73 +90,84 @@ Aggregate across simulated communities (when dataset index is `not` unique)
...
 
-Example for the 78Mbp simulated community
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. code:: bash
+    
+    autometa-download-dataset \
+        --community-type simulated \
+        --community-sizes 78Mbp \
+        --file-names binning.tsv.gz\
```
WiscEvan, line 123: We should also download the `reference_assignments.tsv.gz` here so this section can be performed standalone

```suggestion
        --file-names binning.tsv.gz,reference_assignments.tsv.gz \
```

docs/source/benchmarking.rst
```diff
@@ -71,73 +90,84 @@ Aggregate across simulated communities (when dataset index is `not` unique)
...
 
 .. code:: bash
 
-    file_id="15CB8rmQaHTGy7gWtZedfBJkrwr51bb2y"
-    gdown --id ${file_id} -O metagenome.fna.gz
-    # or
-    gdown https://drive.google.com/uc?id=${file_id} -O metagenome.fna.gz
+    autometa-benchmark \
+       --benchmark clustering \
+       --predictions binner1_78Mbp_output.tsv.gz binner2_78Mbp_output.tsv.gz binner3_78Mbp_output.tsv.gz \
```
WiscEvan, line 148: Why not just list some of the binners outputs from KwanLab/metaBenchmarks?

docs/source/benchmarking.rst
```diff
@@ -71,73 +90,84 @@ Aggregate across simulated communities (when dataset index is `not` unique)
...
+------------------
 
-    Unfortunately, at the moment ``gdown`` doesn't support downloading entire directories from Google drive.
-    There is an open `Pull request <https://github.com/wkentaro/gdown/pull/90#issue-569060398>`_ on the ``gdown`` repository
-    addressing this specific issue which we are keeping a close eye on and will update this documentation when it is merged.
+.. code:: bash
 
+    autometa-benchmark \
+        --benchmark classification \
+        --predictions taxonomic_profiler1_78Mbp.tsv.gz taxonomic_profiler2_78Mbp.tsv.gz taxonomic_profiler3_78Mbp.tsv.gz  \
```
WiscEvan, line 160: Why not list some of the taxon-profilers outputs in KwanLab/metaBenchmarks?

WiscEvan (changes requested):
Comments regarding formatting, headers and some of the associated code blocks

Dockerfile
```diff
@@ -29,7 +29,6 @@ RUN conda install -c bioconda -c conda-forge --file=requirements.txt \
     && conda clean --all -y \
     && python -m pip install trimap
 
-
```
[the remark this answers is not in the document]
Sidduppal: I didn't make any changes. Not sure why it's showing this.

docs/source/benchmarking.rst
Sidduppal: Completely agreed that the headers should be descriptive. However, in my opinion, they should also be a bit concise. Headers are generally short and sweet, gives the user a gist of what's in that section.

docs/source/benchmarking.rst
Sidduppal: See comment above

docs/source/benchmarking.rst
Sidduppal: Hahaha.....Interesting. I personally like the text to be in a single line. 
If I put the text on multiple lines and someone else tries to open it on their end, then depending on their screen size the text may be too much or too little pe line. 
I prefer to wrap the text while writing so that it appears on different lines. So now even though technically everything is on a single like, it is personalized for each user, as wrapping is then done on their end.

docs/source/benchmarking.rst
Sidduppal: I can, I just didn't want people to upset people. Some people might think that we are specifically targeting or pinpointing the above-mentioned binners.

docs/source/benchmarking.rst
Sidduppal: See comment above

docs/source/benchmarking.rst
chasemc: While it's easier to read in a single line with wrapping, it's better to do separate lines so edits appear as less lines/are easier to quickly see in the commit diff.

docs/source/benchmarking.rst
```diff
@@ -71,73 +90,84 @@ Aggregate across simulated communities (when dataset index is `not` unique)
...
---------------------------------------
-
-Autometa is packaged with a built-in module that allows any user to download any of the available test datasets.
-To use these utilities simply run the command ``autometa-download-dataset``.
+Benchmark classification
+^^^^^^^^^^^^^^^^^^^^^^^^
 
-For example, to download all of the simulated communities reference binning/taxonomy assignments as well as the Autometa
-v2.0 binning/taxonomy predictions:
+Previously we benchmarked the clustering step. Now we are going to benchmark the taxon profiling step. The only difference is that the ``--classification`` flag is provide instead of the ``--clustering`` flag.
```
chasemc, line 96: `--classification` and `--clustering` are values in the example below, not flags

chasemc (changes requested):

chasemc:
`Autometa Test Datasets` should be combined with `Downloading Test Datasets` and should be in the same location as `Downloading Test Datasets`.
Titles:
- `Benchmark Datasets`
  - `Downloading Test Datasets`
  - `Explanation of Datasets` 
    - `Synthetic Communities`
    - `Simulated Communities`
    - `Generating New Simulated Communities`

Move `Example benchmarking with simulated communities` to directly before `Benchmark clustering`

chasemc:
I think it may be helpful to throw an example into Jupyter notebook and link to it (or https://docs.readthedocs.io/en/stable/guides/jupyter.html)

docs/source/benchmarking.rst
chasemc: My opinion is that this overly complicated as a  example. I would remove the for loop and analyzing multiple datasets at once. That could be covered in an "Advanced" section.

chasemc:
I think the document would benefit from a bit of rearrangement
Including:
`Autometa Test Datasets` should be combined with `Downloading Test Datasets` and should be in the same location as `Downloading Test Datasets`.
Titles:
- Benchmarking Datasets
  - Download Datasets
   - Explanation of Datasets
     - Synthetic Communities
     - Simulated Communities
     - Generating New Simulated Communities

Move the heading `Example benchmarking with simulated communities` to directly before the heading `Benchmark clustering`

----

Make the examples for each type of clustering clearer by presenting them in the same context :

- `Benchmark clustering` is presented in the context of for-looping over community sizes
- Technically `Benchmark classification` is presented in the context of for-looping over community sizes but the for-loop isn't shown
- `Benchmark clustering-classification`  is presented in the context of running only on a single community size

The example should be simple, so I would suggest presenting only with the context of running on a single sample/community size. 

An "Advanced" section could discuss how to handle running on multiple samples.

----

Generally, the commands/code need more description. As it stands now it's still unclear what commands to run if I want to benchmark (Is following the section `Benchmark clustering-classification` the same as running `Benchmark clustering` and `Benchmark classification` separately?)
Possibly could be done with more descriptive headings (e.g. `Benchmark clustering` becomes `Benchmark Binning Results`; `Benchmark classification` becomes `Benchmark taxonomic assignments`), etc

`Aggregate results across simulated communities` seems to be just data handling? If just data-handling, it makes the documentation less clear; remove or place in an "Advanced" section or similar at end. 

Flags such as `--output-long` and  `--output-classification-reports` aren't described/defined
Inputs aren't described, e.g. how is the input file for  `--predictions`, `--reference`, etc supposed to be structured?

The documentation is missing discussion of what `autometa-benchmark` actually does, and the results it produces.

WiscEvan:
Closing this for now. Please submit a new PR from a new KwanLab/Autometa branch with any changes still necessary.

Addressed a few comments in https://github.com/KwanLab/Autometa/pull/215

PR #201, opened 2021-12-03
:art::snake: Add specific parsers for domtblout and tblout for hmmscan output formats

WiscEvan:
Placing this here in case anyone would like to add their thoughts. `hmmscan` can output multiple formats, for example `output.tblout` and `output.domtblout`. Previously the marker annotations parsed the tblout file format. However, the domtblout format contains more information that could be used to enhance the marker databases/filtering/annotation process. Happy to close this PR without a merge, but thought I would see if there was any interest in including this.


ajlail98 (approved):
These changes looks good to me! I understand what they changed at a high level but I did not actually test run the code. Let me know if you've done that or whether you would be interested in talking through that with me!

Issue #186, opened 2021-09-01
Add read-based contig coverage instead of spades-based


chasemc:
Partially implemented subworkflow needs to be updated to nf-core 2 format
https://github.com/KwanLab/Autometa/blob/8934bf52ae7ab4b91018b0cfc9020f50280fd201/subworkflows/local/contig_coverage.nf

WiscEvan:
What else needs to be updated/implemented for this subworkflow?

chasemc:
Been awhile but I don't think reads are considered at all yet because it was extra complication switching the existing subworkflow to the metamap format.

We'll have to decide on the best input format first. My vote would be to start with tsv: https://nf-co.re/mag/2.1.0/usage#samplesheet-input-file

PR #218, opened 2022-01-18
🍏 Change Nextflow I/O behavior

### LINKED ISSUE #160: Do you think it would be worth changing the I/O behavior here s.t. each input metagenome has its own directory?
Do you think it would be worth changing the I/O behavior here s.t. each input metagenome has its own directory? 

Maybe we change this to output to follow nf-core guidelines (one `--input` arg. and one `--output` arg) to specify where results should be placed.

The proposed `--output` directory structure:

```bash
-- output/
    +-- input_metagenome_1
    │   \-- kmer results related to metagenome_1
    │   \-- binning results related to metagenome_1
    │   \-- ... etc.
    +-- input_metagenome_2
    │   \-- ... files related to metagenome_2
    \-- etc.
```

_Originally posted by @WiscEvan in https://github.com/KwanLab/Autometa/pull/157#r636964912_

WiscEvan:
Nextflow output structure now resembles what was discussed in #160. Note, metagenomes are not enumerated for generation of their output directory name, their `meta.id` is used which is the `metagenome.simpleName` (groovy method) from the respective input metagenome. This clobbers filenames with multiple `.`. For example, the `meta.id` of `my.example.metagenome.fasta` would be `my`. Moving forward (looking at #186), altering the input s.t. a sample sheet is provided would use values in this table to check for unique sample IDs to write each sample to its respective sample ID results directory.

- 🍏  fixes #160
- :memo: Update running Autometa documentation
  - :fire: Remove trailing whitespaces
  - :art: Change some of the note card headers to attention and caution cards
- :fire: Remove redundancies in main autometa workflow
- 🍏  Add `storeDir` for fetching mock_data genomes
- 🍏 🛠️ :bug: `hmmsearch` nf processes still need to be fixed
- :fire: Remove unused parameters in `nextflow_schema.json`
- :fire::art: Fix `nextflow.config` nf-core settings to silence linter warnings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,7 +162,4 @@ pipeline_info
 work/
 large_db/
 nf-params.json
-autometa_interim_dir/
-autometa_outdir/
-autometa_tracedir/
-
+autometa-nxf-output
--- a/docs/source/running-autometa.rst
+++ b/docs/source/running-autometa.rst
@@ -51,7 +51,7 @@ The following workflow is recommended:
 
     If you use SPAdes then Autometa can use the k-mer coverage information in the contig names. If you have used any other assembler, then you first have to make a coverage table.
 
-    Fortunately, Autometa can construct this table for you with: ``autometa-coverage``. Use ``--help`` to get the complete usage. :ref:`coverage`
+    Fortunately, Autometa can construct this table for you with: ``autometa-coverage``. Use ``--help`` to get the complete usage or for a few examples see :ref:`coverage-calculation`
 
 
 Basic
--- a/docs/source/running-autometa.rst
+++ b/docs/source/running-autometa.rst
@@ -13,12 +13,12 @@ Nextflow helps Autometa produce reproducible results while allowing the pipeline
 System Requirements
 ===================
 
-Currently the nextflow pipeline requires Docker so it must be installed on your system. 
+Currently the nextflow pipeline requires Docker so it must be installed on your system.
 If you don't have Docker installed you can install it from `docs.docker.com/get-docker <https://docs.docker.com/get-docker>`_.
-We plan on removing this dependency in future versions, so that other dependency managers 
+We plan on removing this dependency in future versions, so that other dependency managers
 (e.g. Conda, Singularity, etc) can be used.
 
-Nextflow runs on any Posix compatible system. Detailed system requirements 
+Nextflow runs on any Posix compatible system. Detailed system requirements
 can be found in the `nextflow documentation <https://www.nextflow.io/docs/latest/getstarted.html#requirements>`_
 
 Nextflow (required) and nf-core tools (optional but highly recommended) installation will be discussed in :ref:`install-nextflow-nfcore-with-conda`.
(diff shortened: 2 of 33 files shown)


WiscEvan:
Will be merging by EOTD tomorrow unless an issue is raised

chasemc:
Few things to address, I'll probably edit this with more notes...

- storedir for `process PREPARE_LCA` will be wherever nextflow is running from, should be variable 
  - https://github.com/KwanLab/Autometa/blob/3c65a5789bc3ddd9e95340674224372675b89420/modules/local/prepare_lca.nf#L21
  
- because outdir defaults to `baseDir` and the output directory name is dyanmic, there is no way to gitignore the output directory, would have to individual gitignore all output file patterns which seems a bit dangerous
  -https://github.com/KwanLab/Autometa/blob/0cba818136f921289a961008e29e81c50f9d86b0/nextflow.config#L40

chasemc:
> Hardcoding the outdir would be a simple fix, right? Should we go this route?

Yeah I think that would be fine. 


> I'm not sure what you mean by storeDir being variable. Do you mean something like caching these precomputed dbs to params.outdir?

Right now it's hardcoded and so will always be created based on where nextflow is run from 
Should be a parameter, could maybe default to a folder under params.outdir

As an aside, how is the LCA stuff kept in check with the nr db?  If someone downloads the new nr.gz, `PREPARE_LCA` should be aware right? May have to keep track of file hashes

WiscEvan:
> Should probably add a check at the start of the pipeline and fast fail if `$outdir/whatever` isn't empty?

I'm not convinced this is necessary at the moment. This could either be a job for nextflow or could cause some problems if the end-user is not using nextflow properly. We'll kick the can down the road for now.

WiscEvan:
> > I'm not sure what you mean by storeDir being variable. Do you mean something like caching these precomputed dbs to params.outdir?
> 
> Right now it's hardcoded and so will always be created based on where nextflow is run from Should be a parameter, could maybe default to a folder under params.outdir
> 
> As an aside, how is the LCA stuff kept in check with the nr db? If someone downloads a new nr.gz, `PREPARE_LCA` should be aware right? May have to keep track of file hashes

I think this is _as intended_ for wherever nextflow is run b/c these LCA dbs can be used across runs of different datasets. If I am recalling correctly, if the ncbi databases change, the cached LCA databases will be regenerated (I think nextflow is already doing some of this file hash tracking behind the scenes here).

chasemc:
> > Should probably add a check at the start of the pipeline and fast fail if `$outdir/whatever` isn't empty?
> 
> I'm not convinced this is necessary at the moment. This could either be a job for nextflow or could cause some problems if the end-user is not using nextflow properly. We'll kick the can down the road for now.

IMO- Because this PR removes the provenance (run ID) from the output I think either the pipeline should fail if files already exist or the provenance should be written as an output. This may be a larger issue but this PR does take a step backwards in data  provenance

WiscEvan:
Happy to add that in for the 2.1.0 release 👍

PR #220, opened 2022-01-19
Update mock_data_reporter.Dockerfile

chasemc:
Couldn't get previous dockerfile to work (also R package installation from conda can be tricky)
Updated Dockerfile to reproducible build using versioned Rocker base image and R package downloads from MRAN


WiscEvan:
Nice! Will put this in as this is the only failing docker build for the docker CI at the moment. The current dockerfile was working on both deep-thought and kwanlab server, but was not building on my Mac. Thanks for the fix 👍 🛠️

Issue #177, opened 2021-07-02
Add nf-core licenses to imported modules


chasemc:
Modules installed via nf-core don't currently transfer licenses over (see 
https://github.com/nf-core/tools/issues/1151). Add the nf-core modules licences to these.

WiscEvan:
Following https://nf-co.re/developers/tutorials/unofficial_pipelines#acknowledging-nf-core, the section mentioned was placed in the README. 👌

PR #222, opened 2022-01-19
Simplify Licensing

chasemc:
We discussed in a group meeting the option of pointing to LICENSE instead of keeping track across 88383838+ files. This is what that looks like. 

Still doesn't address #177


WiscEvan (approved):
👍

PR #228, opened 2022-01-25
Refactor samplesheet

WiscEvan:
Now samplesheet takes input `cov_from_assembly` instead of `cov_from_contig_headers` so in the future, other assembler contig header parsers may be easily added.


modules/local/align_reads.nf
```diff
@@ -29,7 +29,7 @@ process ALIGN_READS {
         path "*.version.txt"                   , emit: version
 
     when:
-        !meta.cov_from_spades
```
chasemc, line 32: I think these `when` control flows should be handled within the calling workflow/subworkflow because I don't think it's unreasonable to think these modules might be reused for other subworkflows in the future. It also will probably allow defining that flow in a single place

WiscEvan: I agree. Although I'm not sure the `when` directive can be used within a workflow

PR #253, opened 2022-04-05
:green_apple::snake::art: add exit status code of 204 for raised BinningError

WiscEvan:
- :snake: Add exit status code of `sys.exit(204)` for `BinningError('no clusters recovered')` in `autometa-binning`
- :snake::memo: Add extra warnings when this error is reached (i.e. Mentions as to why no clusters may have been recovered from this dataset and for the user to inspect their inputs).
- :green_apple::bug: Previously if no clusters were recovered from a particular dataset, the entire workflow would terminate, even though this dataset should just be ignored for downstream processes.

- :green_apple: `binning.nf` module already ignores `204` exit status as this is 'no content', continuing with the other input datasets but ignoring this for downstream processes.

### DIFF
--- a/autometa/binning/recursive_dbscan.py
+++ b/autometa/binning/recursive_dbscan.py
@@ -888,19 +888,30 @@ def main():
     logger.info(f"Selected clustering method: {args.clustering_method}")
 
     if args.taxonomy:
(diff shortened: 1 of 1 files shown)


Sidduppal (approved):
Looks good.

PR #254, opened 2022-04-05
:art::bug::snake: Convert 'unclassified' taxids of 0 to root taxid of 1

WiscEvan:
:art::snake: Some taxon-profilers assign unclassified contigs to a taxid of 0 where others assign root taxid of 1.

The NCBI class method of converting the taxid datatype expects a positive integer, so we are converting the unclassified taxid of 0 to the root taxid of 1. All input predictions will be converted in this same way. The contigs with their converted taxids as well as the number of 'unclassified' taxids are now emitted in the logs during conversion.

### DIFF
--- a/autometa/validation/benchmark.py
+++ b/autometa/validation/benchmark.py
@@ -228,6 +228,11 @@ def get_target_labels(
     pred_df = pd.read_csv(
         prediction, sep="\t", index_col="contig", usecols=["contig", "taxid"]
     ).convert_dtypes()
+    # Convert taxids of 0 to 1 (some taxon-profilers assign unclassified to 0)
+    logger.debug(pred_df[pred_df.taxid.eq(0)].index.unique().tolist())
+    logger.debug(f"Converting {pred_df.taxid.eq(0).sum():,} taxids from 0 to 1")
+    pred_df.taxid = pred_df.taxid.map(lambda tid: 1 if tid == 0 else tid)
+
     if not isinstance(reference, pd.DataFrame) and isinstance(reference, str):
         ref_df = (
             pd.read_csv(


chasemc:
Would probably use something like:
'pred_df.taxid = pred_df.taxid.clip(lower=1)'

WiscEvan:
> May be cleaner: `pred_df.taxid = pred_df.taxid.clip(lower=1)`
> 
> could also be done inplace

Nice, I was unaware of this method. 

I'm not sure if this is preferred or more understandable 🤷 

Or if `inplace` modification vs. reassignment is preferred.

Sidduppal:
Both ways look good to me. You can chose and I can approve it.

WiscEvan:
The only concern I have is whether this will work if any of the taxids are strings rather than a `float` or `int`?

chasemc:
> The only concern I have is whether this will work if any of the taxids are strings rather than a `float` or `int`?

it would error

WiscEvan:
Ah, yeah. 

```python
df.taxid.clip(lower=1)
TypeError: '>=' not supported between instances of 'str' and 'int'
```

The current mapping method will leave any 'str' types to be converted with `ncbi.convert_taxid_dtype(...)`

WiscEvan:
@Sidduppal, then I think this is ready for review. Albeit maybe not as clean, it leaves the exceptions for `ncbi.convert_taxid_dtype`

chasemc (approved):
```

import pandas as pd
pred_df = pd.DataFrame({'a': [0, -1, 2], 'taxid': [0, -1, 2]})
pred_df.taxid = pred_df.taxid.map(lambda tid: 1 if tid == 0 else tid)
pred_df

   a  taxid
0  0      1
1 -1     -1
2  2      2
```

PR #265, opened 2022-04-19
:snake::bug: Fix parsing taxids to delnodes set

WiscEvan:
:snake::bug: delnodes.dmp taxids were parsed and added to the `self.delnodes` set. However these taxids were not converted to integers prior to being stored within the set, causing a check of a current taxid (an integer) to return `False` due to the taxid's string representation being stored.


Sidduppal (approved):
Nice to know you found it!

PR #294, opened 2022-09-14
test pull request for myst-formatted documentation

kaw97:
pull request for tests/troubleshooting


WiscEvan (commented):
Is the `rst` directory necessary? If so, this should be a sub-directory within the `docs` directory. Also, is the Dockerfile necessary?

PR #316, opened 2023-03-27
singularity image urls

kaw97:
made new branch to work from for singularity URL hard-coding


chasemc (approved):
Looks good.

Note for any other reviewers- A couple of conda calls are different than the docker/singularity but that's not what was being worked on here and I've changed them in my update so it's fine to leave as is

chasemc:
Only note is I've never used singularity so haven't tested this but assume that since the link is valid it should work

PR #320, opened 2023-04-03
singularity image urls

kaw97:
hard-coded URLs for singularity images


kaw97:
forgot I opened one already

Issue #328, opened 2023-06-15
autometa-update-databases, error building GTDB diamond database


Kohtzanth:
# Current Behavior

The GTDB gtdb_proteins_aa_reps.tar.gz and taxdump files download correctly, but an error occurs when trying to build the diamond database. After the error there is an empty gtdb.faa file, so I think the files are not being merged after the extraction step.

## Steps to Reproduce

```bash
autometa-setup-gtdb --reps-faa mambaforge/envs/autometa/database/gtdb_proteins_aa_reps.tar.gz --dbdir mambaforge/envs/autometa/database --cpus 80
[06/14/2023 07:52:49 PM DEBUG] autometa.taxonomy.gtdb: Extracting tarball containing GTDB ref genome animo acid data sequences to: mambaforge/envs/autometa/database/protein_faa_reps
[06/14/2023 07:56:26 PM DEBUG] autometa.taxonomy.gtdb: Extraction done.
[06/14/2023 07:56:26 PM DEBUG] autometa.taxonomy.gtdb: Merging 0 faa files.
[06/14/2023 07:56:26 PM DEBUG] autometa.taxonomy.gtdb: Combined GTDB faa file written to mambaforge/envs/autometa/database/gtdb.faa
[06/14/2023 07:56:27 PM DEBUG] autometa.common.external.diamond: diamond makedb --in mambaforge/envs/autometa/database/gtdb.faa --db mambaforge/envs/autometa/database/gtdb.dmnd -p 80
Traceback (most recent call last):
  File "/home/anthonyk/mambaforge/envs/autometa/bin/autometa-setup-gtdb", line 10, in <module>
    sys.exit(main())
  File "/home/anthonyk/mambaforge/envs/autometa/lib/python3.9/site-packages/autometa/taxonomy/gtdb.py", line 356, in main
    diamond.makedatabase(
  File "/home/anthonyk/mambaforge/envs/autometa/lib/python3.9/site-packages/autometa/common/external/diamond.py", line 51, in makedatabase
    subprocess.run(
  File "/home/anthonyk/mambaforge/envs/autometa/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['diamond', 'makedb', '--in', 'mambaforge/envs/autometa/database/gtdb.faa', '--db', 'mambaforge/envs/autometa/database/gtdb.dmnd', '-p', '80']' returned non-zero exit status 1.

```

[...]

PR #329, opened 2023-06-16
:bug: Fix GTDB database setup

Sidduppal:
:bug: Fix bug with running GTDB taxonomic workflow
🐛  Fix bug with setting up gtdb database. [issue328](https://github.com/KwanLab/Autometa/issues/328)


chasemc:
Why do the files need to be decompressed?

Sidduppal:
Decompression is required to modify the fasta headers and then concatenate the sequences together for the final database.

chasemc:
That's the .sh files /grep that was also modified?

Sidduppal:
Yes, that's an independent bug fix that I found during some testing.

chasemc:
Possible to use zgrep on still gzip'ed files instead?

jason-c-kwan:
Or pipe the output from zcat, or use the gzip module in Python

Sidduppal:
@chasemc 
> Possible to use zgrep on still gzip'ed files instead?

We are doing file modification on the files to change the FASTA headers. `zgrep` would just get the header but modifying it would be hard if it's not unzipped. It was also require needing to use external subprocesses and not internal python modules.

@jason-c-kwan 
> Or pipe the output from zcat, or use the gzip module in Python

I am currently using the `gzip` module in python for file manipulation.
I believe a possible scenario could be there where you use `zcat` modify the header and then it, but I don't think how efficient it would be as compared to unzipping the files which takes around 5-10 min.

chasemc:
My confusion and suggestion of zgrep was because I thought the edits were related (in the future try to only fix a single thing in a PR or at least separate out the commits)

My question is then the same as Jason's- is there a reason not to just read the files using the gzip module rather than decompress, write and then read back in

Is the following code's single purpose to read some fasta files, edit the identifier and then concatenate into a single file?
https://github.com/KwanLab/Autometa/blob/255066a2cdd9ed9371a2b68a344a269adee56554/autometa/taxonomy/gtdb.py#L57C2-L103

jason-c-kwan:
Yeah seems like there are too many steps.
1. Get protein accession from filepath (can be done on gz)
2. Open combined gz file for writing with gzip module
3. Open each component file with gzip, write line to output gzip, change header line as appropriate
4. Close output file.
Not above, all files can remain gzipped, but you are effectively copying all the input faa files into a combined file. Is this necessary?

chasemc:
Just to comment before I leave for the weekend...
If the answer is yes, if possible, probably best to read the desired files (filename match) directly from the tar, edit the header/id while reading and write directly into the concatenated file. Note: I'm not familiar with this section of code function and I don't know the structure of the tar file so this may or may not be a good suggestion

chasemc:
Hit submit before seeing @jason-c-kwan  responded

autometa/taxonomy/gtdb.py
```diff
@@ -57,9 +59,9 @@ def create_gtdb_db(reps_faa: str, dbdir: str) -> str:
         tar.close()
         logger.debug("Extraction done.")
         reps_faa = dbdir
-
+ 
     genome_protein_faa_filepaths = glob.glob(
-        os.path.join(reps_faa, "**", "*_protein.faa"), recursive=True
+        os.path.join(reps_faa, "**", "*_protein.faa.gz"), recursive=True
```
WiscEvan, line 64: Should this also find files that the user may have decompressed? Are these files still decompressed in earlier database versions? If so, we can provide backwards compatibility here

 ```suggestion
        os.path.join(reps_faa, "**", "*_protein.faa*"), recursive=True
        # To find *_protein.faa and *_protein.faa.gz files
```

Perhaps there is a literal glob character to optionally find `[.gz]` files instead of wildcard? 🤷 

```python
        os.path.join(reps_faa, "**", "*_protein.faa[.gz]"), recursive=True # I don't think this works?
```

autometa/taxonomy/gtdb.py
```diff
@@ -76,17 +78,20 @@ def create_gtdb_db(reps_faa: str, dbdir: str) -> str:
...
         for faa_file, acc in faa_index.items():
-            with open(faa_file) as f_in:
+            with gzip.open(faa_file, "rb") as f_in:
                 for line in f_in:
-                    if line.startswith(">"):
-                        seqheader = line.lstrip(">")
+                    # print(line.decode('utf-8'))
+                    if line.decode('utf-8').startswith(">"):
+                        seqheader = line.decode('utf-8').lstrip(">")
                         line = f"\n>{acc} {seqheader}"
```
WiscEvan, line 89: - :fire: print statement
- :art: Can remove redundancy and times `str.decode('utf-8')` is called (DRY) with a single line
- :art: Perhaps a different variable for the line being read vs. written could improve readability?

> Some ternary operators may allow handling both compressed and decompressed files here

```suggestion
                    line = line.decode("utf-8")
                    if line.startswith(">"):
                        seqheader = line.lstrip(">")
                        outline = f"\n>{acc} {seqheader}"
```

autometa/taxonomy/gtdb.py
```diff
@@ -76,17 +78,20 @@ def create_gtdb_db(reps_faa: str, dbdir: str) -> str:
...
-                    f_out.write(line)
-    logger.debug(f"Combined GTDB faa file written to {combined_faa}")
-    return combined_faa
+                        f_out.write(line)
+                    else:    
+                        f_out.write(line.decode('utf-8'))
```
WiscEvan, line 92: Following suggestions on lines `86+`-`89+`

```suggestion
                    else:    
                        outline = line
                    f_out.write(outline)
```

workflows/autometa.sh
```diff
@@ -231,10 +231,10 @@ then
         set -x
         grep ">" $kingdom_fasta | \
             sed 's/^>//' | \
-            sed 's/$/_/' | \
-            cut -f1 -d" " > $orf_prefixes
+            cut -f1 -d" " | \
+            sed 's/$/_/' > $orf_prefixes
         # Retrieve ORF IDs from contig IDs
-        grep -f $orf_prefixes $orfs | sed 's/^>//' | cut -f1 -d" " > $orf_ids
```
WiscEvan, line 237: I think this should be moved to a different PR. This is not relevant to the database setup correct?

WiscEvan (changes requested):
It looks like some formatting should be done prior to this being merged (details in the comments). Otherwise I had just a few questions & suggestions.

autometa/taxonomy/gtdb.py
```diff
@@ -76,17 +78,20 @@ def create_gtdb_db(reps_faa: str, dbdir: str) -> str:
         os.makedirs(dbdir)
 
     logger.debug(f"Merging {len(faa_index):,} faa files.")
-    combined_faa = os.path.join(dbdir, "gtdb.faa")
+    combined_faa = os.path.join(dbdir, "gtdb.faa") 
     with open(combined_faa, "w") as f_out:
         for faa_file, acc in faa_index.items():
-            with open(faa_file) as f_in:
+            with gzip.open(faa_file, "rb") as f_in:
```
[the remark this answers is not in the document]
Sidduppal: No, but we are assuming someone will only use this script when they need to setup the databases. So you download the amino acid files each sequence is gunzipped and if a user untars that tarball (without unzipping each sequence) we can handle that too.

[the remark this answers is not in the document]
WiscEvan: :+1:

WiscEvan (approved):
👍

bheimbu:
Hi,

I'd like to use the gtdb database, but I'm not able to build it. Any news when this will be fixed?

Cheers Bastian

chasemc:
Rolling back cython as suggested by some comments in 

> Looks like the tests are failing due to a recent issue with hdbscan and cython ([scikit-learn-contrib/hdbscan#600](https://github.com/scikit-learn-contrib/hdbscan/issues/600))

didn't work. .

PR #339, opened 2023-08-09
:bug::shell: Fix GTDB taxon-binning workflow

shaneroesemann:
Implement changes by @Sidduppal in issue https://github.com/KwanLab/Autometa/pull/329#issuecomment-1594996732

### DIFF
--- a/workflows/autometa-large-data-mode.sh
+++ b/workflows/autometa-large-data-mode.sh
@@ -244,7 +244,7 @@ then
             cut -f1 -d" " | \
             sed 's/$/_/' > $orf_prefixes
         # Retrieve ORF IDs from contig IDs
-        grep -f $orf_prefixes $orfs | cut -f1 -d" " | sed 's/^>//'  > $orf_ids  
+        grep -f $orf_prefixes $orfs | cut -f1 -d" " | sed 's/^>//' > $orf_ids
         # Retrieve ORF seqs from ORF IDs
         seqkit grep \
             --pattern-file $orf_ids \
--- a/workflows/autometa-large-data-mode.sh
+++ b/workflows/autometa-large-data-mode.sh
@@ -241,10 +241,10 @@ then
         set -x
         grep ">" $kingdom_fasta | \
             sed 's/^>//' | \
-            sed 's/$/_/' | \
-            cut -f1 -d" " > $orf_prefixes
+            cut -f1 -d" " | \
+            sed 's/$/_/' > $orf_prefixes
         # Retrieve ORF IDs from contig IDs
-        grep -f $orf_prefixes $orfs | sed 's/^>//' | cut -f1 -d" " > $orf_ids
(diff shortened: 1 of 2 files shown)


Sidduppal (approved):
Looks good.

workflows/autometa-large-data-mode.sh
```diff
@@ -241,10 +241,10 @@ then
         set -x
```
WiscEvan, line 241: Would you mind adding a comment of an example header?

```suggestion
        # <insert example ORF header>
        set -x
```

workflows/autometa.sh
```diff
@@ -231,10 +231,10 @@ then
         set -x
```
WiscEvan, line 231: Would you mind adding a comment of an example header?

```suggestion
        # <insert example ORF header>
        set -x
```

workflows/autometa-large-data-mode.sh
```diff
@@ -241,10 +241,10 @@ then
...
         grep ">" $kingdom_fasta | \
             sed 's/^>//' | \
-            sed 's/$/_/' | \
-            cut -f1 -d" " > $orf_prefixes
+            cut -f1 -d" " | \
+            sed 's/$/_/' > $orf_prefixes
         # Retrieve ORF IDs from contig IDs
-        grep -f $orf_prefixes $orfs | sed 's/^>//' | cut -f1 -d" " > $orf_ids
+        grep -f $orf_prefixes $orfs | cut -f1 -d" " | sed 's/^>//'  > $orf_ids
```
WiscEvan, line 247: Why must the order of `sed` and `cut` be swapped in this line?

:art::fire: Removing unnecessary whitespace

```suggestion
        grep -f $orf_prefixes $orfs | cut -f1 -d" " | sed 's/^>//' > $orf_ids
```

workflows/autometa.sh
```diff
@@ -231,10 +231,10 @@ then
...
         grep ">" $kingdom_fasta | \
             sed 's/^>//' | \
-            sed 's/$/_/' | \
-            cut -f1 -d" " > $orf_prefixes
+            cut -f1 -d" " | \
+            sed 's/$/_/' > $orf_prefixes
         # Retrieve ORF IDs from contig IDs
-        grep -f $orf_prefixes $orfs | sed 's/^>//' | cut -f1 -d" " > $orf_ids
+        grep -f $orf_prefixes $orfs | cut -f1 -d" " | sed 's/^>//'  > $orf_ids
```
WiscEvan, line 237: Why must the order of sed and cut be swapped in this line?

WiscEvan (commented):
Would you mind updating the description for how these changes resolve the issue? Currently this is somewhat unclear. Not saying that the issue has not been resolved, it is just difficult to discern with the information that is currently present. Is there a specific instance where partial matching was causing a problem?

workflows/autometa.sh
chasemc: Probably doesn't matter much but might be faster since there's less data being piped to the last step

Issue #350, opened 2023-11-12
line_terminator not lineterminator use in .to_csv from pandas throws error with pandas > 1.5


hyphaltip:
autometa-binning-summary throw error because of different expectation for .to_csv - I assume this is a pandas version problem I have? per https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html which changed after 1.5.0 but the env requirements state only pandas > 1.1 (https://github.com/KwanLab/Autometa/blob/main/autometa-env.yml) 

```
+ autometa-binning-summary --binning-main results_autometa_mh/Boulby_Sample_1/autometa/Boulby_Sample_1.gtdb.archaea.dbscan.main.tsv --markers results_autometa_mh/Boulby_Sample_1/autometa/Boulby_Sample_1.archaea.markers.tsv --metagenome results_megahit/Boulby_Sample_1/final.contigs.fa --dbdir /srv/projects/db/autometa
/2.0/gtdb/ --dbtype gtdb --output-stats results_autometa_mh/Boulby_Sample_1/autometa/Boulby_Sample_1.gtdb_archaea_metabin_stats.tsv --output-taxonomy results_autometa_mh/Boulby_Sample_1/autometa/Boulby_Sample_1.gtdb_archaea_metabin_taxonomy.tsv --output-metabins results_autometa_mh/Boulby_Sample_1/autometa/Boulby_Sam
ple_1.gtdb_archaea_metabins
[11/12/2023 02:18:59 PM INFO] autometa.binning.summary: Retrieving metabins' stats for cluster
[11/12/2023 02:18:59 PM INFO] root: Wrote metabin stats to results_autometa_mh/Boulby_Sample_1/autometa/Boulby_Sample_1.gtdb_archaea_metabin_stats.tsv
[11/12/2023 02:19:01 PM DEBUG] autometa.taxonomy.gtdb: Processing nodes from /srv/projects/db/autometa/2.0/gtdb/merged.dmp
[11/12/2023 02:19:01 PM DEBUG] autometa.taxonomy.gtdb: merged loaded
[11/12/2023 02:19:01 PM DEBUG] autometa.taxonomy.gtdb: Processing delnodes from /srv/projects/db/autometa/2.0/gtdb/delnodes.dmp
[11/12/2023 02:19:01 PM DEBUG] autometa.taxonomy.gtdb: delnodes loaded
[...]

PR #353, opened 2023-11-30
Fix Issue#350

Sidduppal:
- Fixes pandas version to `>=1.5`. 
- Fixes issue #350


chasemc:
The failing tests need to be fixed via updating to the new Biopython for gc calculations (note the factor difference (https://github.com/biopython/biopython/issues/4622#issuecomment-1934396369)):

"Function 'GC' in Bio.SeqUtils was deprecated in Release 1.80, and removed in Release 1.82. Instead use function 'gc_fraction'"


https://github.com/biopython/biopython/blob/biopython-183/DEPRECATED.rst#biosequtils

tests/unit_tests/test_metagenome.py
chasemc, line 1: May need to address this later but will merge for now

